PR #91, opened 2026-05-30
docs(v0): in-repo doc/ reference ported from the legacy site

anilcancakir:
## What

Adds an in-repo `doc/` reference to the frozen `0.0.x` line, porting the legacy site documentation (`wind.fluttersdk.com/resources/contents/docs/0.0.3`) into the 1.0.x doc-system format. 52 markdown files, mirroring the 0.0.3 topic taxonomy.

## Why

The 1.0.x line now keeps its docs in-repo under `doc/`. This brings the v0 line to the same shape so the website can sync v0 docs from the branch instead of a separate content store.

## Scope

Docs only. No source under `lib/` or `example/lib/` is touched. Content is faithful to the v0 API (static `WindTheme` class, widgets `WContainer`/`WCard`/`WFlex`/`WFlexContainer`/`WFlexible`/`WText`); format mirrors the master 1.0.x doc system (H1, ToC, anchors, `<x-preview>`, Props/Constructor for widgets, Related Documentation).

## Layout

- `getting-started/` (2), `concepts/` (4), `customization/` (9), `widgets/` (6)
- `typography/` (9), `flex/` (8), `spacing/` (2), `sizing/` (4)
- `backgrounds/` (1), `borders/` (3), `effects/` (1), `layout/` (2), `example/` (1)

## Guarantees

- No 1.0-only API leakage (`WDiv`, `WindThemeData`, builder pattern, `ring-`/`grid-cols`/`animate-`, `toggleTheme`).
- Every `<x-preview source>` resolves to a real `example/lib/pages/**` page on this branch; omitted (not faked) where no page exists (getting-started, customization, wcard, min/max sizing).
- All cross-doc links are relative `.md`; no absolute `wind.fluttersdk.com` URLs.
- Widget Constructor/Props tables read from the actual v0 component sources.
- v0 color shades 50-900 (no 950); barrel import `package:fluttersdk_wind/fluttersdk_wind.dart`.

## Follow-up (not in this PR)

- `lib/src/parsers/font_style.dart` has a trailing-tab typo in a map key (`'not-italic\t'`) that would stop the `not-italic` class from matching at runtime. Documented as `not-italic` here; worth a separate v0 source fix.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,4 @@ migrate_working_dir/
 .flutter-plugins
 .flutter-plugins-dependencies
 build/
+.ac/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 # Changelog
 This project follows [Semantic Versioning 2.0.0](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+- Add in-repo `doc/` reference for the 0.0.x line, ported from the legacy site docs into the 1.0.x doc-system format.
+
 ## [0.0.4] - 2025-06-12
 
 - Updated `platform_info` dependency to `^5.0.0` for improved platform detection.
--- a/doc/backgrounds/background-color.md
+++ b/doc/backgrounds/background-color.md
@@ -1,0 +1,155 @@
+# Background Color
+
+Apply background colors to widgets using predefined theme colors or arbitrary hex values.
+
+- [Basic Usage](#basic-usage)
+- [Quick Reference](#quick-reference)
+- [Arbitrary Values](#arbitrary-values)
+- [Responsive Design](#responsive-design)
+- [Dark Mode](#dark-mode)
+- [Customizing Theme](#customizing-theme)
+- [Related Documentation](#related-documentation)
+
+<x-preview path="backgrounds/background_color" size="md" source="example/lib/pages/backgrounds/background_color.dart"></x-preview>
+
+```dart
+WContainer(
+  className: 'bg-blue-500 w-64 h-64 alignment-center',
+  child: WText('Blue Background', className: 'text-white'),
+)
+```
+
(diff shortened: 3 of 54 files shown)


anilcancakir:
Thanks @copilot. Addressed in 0c652e7:

**Fixed (8):**
- `customization/font-size.md`: font-size utilities now use `text-*` (`text-xs`, `text-3xl`, `text-[8]`); `font-*` is font weight. Verified against `lib/src/parsers/font_size_parser.dart:11` (regex matches `text-`). Added a note disambiguating the two prefixes.
- `typography/font-size.md` + `customization/rounded-corners.md`: formula now reads `value × rem factor (16 by default)` (one multiplication), matching `WindTheme.getRemFactor()` (= pixelFactor × 4) and `pixel-factor.md`. Dropped the misleading `× 4 × 4` framing.
- `widgets/wflexible.md`: clarified the `flex` fallback wording ("neither the `flex` parameter nor a `flex-x` utility is set").
- `widgets/wflexcontainer.md`: `mainAxisSize` now says `axis-min` / `axis-max` instead of `axis-*`.
- `effects/shadow.md`, `borders/border-width.md`, `borders/border-radius.md`: arbitrary-value placeholder is now single-bracket `[<value>]`.

**Declined (2), with source evidence:**
- `typography/letter-spacing.md` "Default px" column and the `0.15 × 4 = 0.6px` example: the parser **does** multiply the stored value by the pixel factor. See `lib/src/parsers/letter_spacing_parser.dart:18` and `:40`: `letterSpacing = WindTheme.getLetterSpacing(size) * WindTheme.getPixelFactor()` (pixel factor = 4 by default). So `tracking-tighter` (-0.05) → -0.2 logical px is correct, and the customization example is correct. The customization page describes the stored em value; the typography page describes the applied pixel value. Both are accurate, so these are kept as-is.

PR #99, opened 2026-06-09
feat(w-div): smart column stretch covers WAnchor and WButton rows

anilcancakir:
## Summary

In a `flex flex-col` with no explicit `items-*`, Wind's smart column cross-axis stretch now covers `WAnchor` (any child) and `WButton` children, not only bare `WDiv`. A clickable nav row authored as `WAnchor(onTap) > WDiv(...)` (WDiv carries no `onTap`) previously rendered at content width; it now fills the column width by default, matching CSS `align-items: stretch` where element type is irrelevant. No explicit `items-stretch` / `w-full` needed.

This continues #97's smart-stretch work, closing the gap the dashboard demo surfaced (nav highlight hugged the text instead of spanning the sidebar).

## Behavior

- `WButton` is stretch-eligible, gated by its own `className`.
- `WAnchor` delegates eligibility to its effective child: inner `WDiv`'s className decides; a `WText` / raw child always stretches.
- Left untouched: explicit width (`w-*` / `min-w-*` / `max-w-*` / `w-full`), self-flex (`grow` / `flex-grow` / `flex-auto` / `flex-initial` / `shrink` / `flex-shrink` / `flex-N`), `absolute`, bare `WText` leaves, raw Flutter widgets.
- Stretch stays `SizedBox(width: double.infinity)` under the existing bounded-width `LayoutBuilder` guard. No `CrossAxisAlignment.stretch` introduced (crash-safe).

## Tests (TDD red-green)

New `test/widgets/w_div/flex_stretch_test.dart` group: WAnchor>WDiv / WAnchor>WText / WButton stretch; explicit-width inner WDiv not stretched; unbounded-width column does not crash; bare WText leaf still excluded.

## Verification

- `dart analyze` 0 issues, `dart format` no diff
- `flutter test` 1381 pass (1 pre-existing skip)
- `./tool/coverage.sh 90` -> 90.6%
- `flutter pub publish --dry-run` clean (post-commit)
- 5-surface sync: `doc/layout/flexbox.md`, `skills/wind-ui/references/{layouts,tailwind-divergence}.md`, `CHANGELOG.md`, `example/lib/pages/layout/flex_grow.dart`; README justified-skip (internal eligibility refinement).

Planned + executed via `/ac:plan` -> `/ac:execute`; reviewer APPROVED.

## Summary by CodeRabbit

* **New Features**
  * Smart column cross-axis stretch now also applies to clickable row containers, so nav/button rows fill column width by default.

* **Documentation**
  * Expanded docs to clarify when auto-stretch applies and explicit exceptions (explicit widths, flex-grow wrappers, absolute children, bare text/raw widgets); rows remain unaffected.
  * Added example demonstrating clickable-row stretching.

* **Tests**
  * Added tests covering stretch behavior and regression scenarios for interactive wrappers.


anilcancakir:
Addressed the review:

- **CodeRabbit (dark: pairs, flex_grow.dart)**: fixed. Every `text-white` in the new stretch demo now carries `dark:text-white` (5 occurrences).
- **Copilot (doc comments imply `WAnchor > WDiv(grow)` is safe/supported, w_div.dart:853/858, flexbox.md:211)**: fixed. Reworded so the comments no longer suggest `grow` inside `WAnchor` is meaningful. `WAnchor > WDiv(grow)` is now called out as unsupported in its own right (the inner `WDiv`'s `Expanded` asserts under the `WAnchor`'s non-`Flex` wrappers regardless of stretch); the self-flex exclusion is described as mirroring the direct-`WDiv` rule (where it genuinely prevents the stretch-compounding assertion), not as a rescue for the wrapped-grow case.
- **Copilot (raw-widget "no width to honor", w_div.dart:872)**: fixed. Reworded to a policy statement: a `WAnchor`-wrapped non-`WDiv` child is stretched so the tap surface fills the column; any explicit width on the wrapped widget still constrains its own content.
- **Copilot (test comment "WButton self-wraps in Expanded", flex_stretch_test.dart:419)**: fixed. WButton does not wrap in Expanded; the comment now says the `grow` token is excluded by `_selfWrapsInFlex` for consistency with the direct-WDiv rule.

- **Copilot (circular import w_div.dart ↔ w_button.dart, line 11)**: keeping as-is. Intra-package file imports that form a cycle are idiomatic in Dart and Flutter (the SDK's own `material` / `widgets` libraries do this pervasively); Dart resolves them and `dart analyze` raises no warning. The "circular dependency" concern applies to circular dependencies between published packages, not file-level imports within one library. Removing it would mean dropping the `child is WButton` eligibility (a locked requirement) for no real benefit.

Verification after the fixes: `dart analyze` 0 issues, `dart format` no diff, `flutter test` green, 0 em-dash in the diff.

PR #105, opened 2026-06-16
ci: auto-create GitHub Release on version tag push

anilcancakir:
## Why

The `publish.yml` workflow publishes to pub.dev on a version-tag push but never created a GitHub Release. As a result `1.0.0` (and earlier `0.0.5`) shipped to pub.dev with no matching GitHub Release; the alpha releases were created by hand. This closes that gap so it cannot be forgotten again.

## What

- New `github-release` job, `needs: publish` and gated on `startsWith(github.ref, 'refs/tags/')`.
- Extracts the matching `## [VERSION]` section from `CHANGELOG.md` as release notes (stops at the next `## ` heading, so `### Added`/`### Fixed` subsections are included).
- Creates the release with `gh release create --verify-tag`. Tags with a hyphen (pre-releases) get `--prerelease`; stable tags get `--latest`.
- Scoped `contents: write` permission on the job only.

## Notes

- The missing `1.0.0` release was created manually as part of this work: https://github.com/fluttersdk/wind/releases/tag/1.0.0
- pub.dev `1.0.0` publish itself succeeded (run 27214539725); nothing was broken there.
- No `lib/` change, so the five-surface post-change sync does not apply.

## Summary by CodeRabbit

* **Chores**
  * Enhanced release publishing process with automated GitHub release creation upon tag push, including proper changelog extraction and version classification.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -18,10 +18,12 @@ jobs:
     name: Validate
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
+        with:
+          persist-credentials: false
 
       - name: Setup Flutter
-        uses: subosito/flutter-action@v2
+        uses: subosito/flutter-action@1a449444c387b1966244ae4d4f8c696479add0b2 # v2
         with:
           channel: stable
           cache: true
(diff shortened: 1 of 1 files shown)


anilcancakir:
Addressed the review findings in f4d09a6:

- **zizmor/unpinned-uses + PinnedDependenciesID** (lines 21/24/51): pinned all actions to commit SHA with a trailing version comment: `actions/checkout@df4cb1c` (v6), `subosito/flutter-action@1a44944` (v2), and the `dart-lang/setup-dart/.github/workflows/publish.yml@65eb853` (v1) reusable workflow. This also clears the pre-existing alerts on the validate and publish jobs, so the whole file is now consistently hash-pinned.
- **zizmor/artipacked** (CodeRabbit duplicate comment): added `persist-credentials: false` to both checkout steps, matching the existing pattern in `zizmor.yml`.
- **Empty release notes** (CodeRabbit): the changelog extraction now captures the section into a variable and fails the job with `::error::No CHANGELOG.md section found for <version>` when the section is missing or whitespace-only, instead of silently creating a release with blank notes. Verified locally both ways (1.0.0 -> 77 lines, 9.9.9 -> guard fails).

Issue #11, opened 2025-12-24
docs: Create Flexbox Layout Example Page


anilcancakir:
## Description
Create a flexbox layout example page showcasing all flex utilities, similar to [tailwindcss.com/docs/flex-direction](https://tailwindcss.com/docs/flex-direction).

## Features to Showcase
- [ ] **Display**: `flex`, `block`
- [ ] **Flex Direction**: `flex-row`, `flex-col` with visual examples
- [ ] **Justify Content**: `justify-start`, `justify-center`, `justify-end`, `justify-between`, `justify-around`, `justify-evenly`
- [ ] **Align Items**: `items-start`, `items-center`, `items-end`, `items-baseline`, `items-stretch`
- [ ] **Align Content**: `align-content-start`, `align-content-center`, etc.
- [ ] **Flex Wrap**: `wrap` display type
- [ ] **Gap**: `gap-{n}`, `gap-x-{n}`, `gap-y-{n}`
- [ ] **Flex Items**: `flex-1`, `flex-auto`, `flex-initial`, `flex-none`, `flex-grow`, `flex-shrink`
- [ ] **Align Self**: `align-self-center`, `align-self-start`, etc.
- [ ] **Axis Size**: `axis-min`, `axis-max`

## Implementation Notes
- Use colored boxes to visualize layout behavior
- Interactive examples if possible
- Show both Row and Column directions

anilcancakir:
Implementation completed in #29. See PR for details.

PR #118, opened 2026-06-24
fix: WPopover opens with interactive triggers and no longer self-dismisses (#11)

anilcancakir:
## What

Fix two WPopover defects:
- **Open failure:** an interactive trigger (WButton/WAnchor) won the gesture arena, so the outer toggle never fired and the popover never opened. Now opens via `Listener(onPointerDown)`, which bypasses the arena.
- **Instant self-dismiss:** the opening tap's pointer-up reached the freshly mounted overlay `TapRegion` as an outside tap, closing on the same frame. Now swallowed by a one-shot flag armed in `open()` and disarmed one frame later (shared `groupId` preserved).

## Tests

New regression set: four behaviors (open, stay-open-on-content-tap, close-on-outside-tap, re-tap-toggles) over both WButton and WDiv triggers, plus the trigger-onTap co-fire and controller-open contracts. Example fixture menu items stretched to full width.

## Verification

`flutter test` 1453 green, `./tool/coverage.sh 90` 91.2%, `dart analyze` + `dart format` clean. Five-surface sync done (doc, skill, CHANGELOG). Verified live in Chrome across all popover example pages.

Fixes REPORT.md #11.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved popover tap behavior so it opens reliably from interactive triggers and no longer dismisses immediately from the same opening tap.
  * Outside taps still close the popover as expected, while taps inside the content keep it open.

* **Documentation**
  * Clarified popover tap behavior and how to use manual control when trigger-on-tap is disabled.

* **Tests**
  * Added regression coverage for trigger opening, inside/outside tap handling, toggle behavior, and controller-driven opening.


anilcancakir:
Pushed 6c5843d addressing the review feedback:

**Accessibility (Copilot + CodeRabbit):** the pointer `Listener` is invisible to assistive tech, so the trigger is now wrapped in `Semantics(button: true, enabled: !disabled, onTap: toggle)`. Screen readers and keyboard activation reach the toggle through the semantic action; pointer input reaches it through the Listener. Added a test that performs `SemanticsAction.tap` and asserts the popover opens.

**Secondary/right-click (Copilot):** the pointer toggle is now filtered to the primary button (`event.buttons == kPrimaryButton`), so a secondary click and scroll-start no longer toggle. Added a test asserting a right-click does not open.

**CHANGELOG (#11) (Copilot):** removed; `#11` was our internal report numbering, not a wind issue. The entry no longer references it and now documents the Semantics/primary-button hardening.

**Coverage (codecov/patch):** added the two tests above. The one remaining uncovered line is the same-frame outside-tap swallow in `_handleTapOutside`, which only fires in the live engine (flutter_test mounts the overlay and runs the post-frame disarm in a single pump, so there is no window where the overlay is mounted and the flag is still armed). Marked `// coverage:ignore-line` with a WHY note per the repo coverage policy; the disarm and close paths are covered.

`flutter analyze` clean, `dart format` no diff, full suite green (1455), coverage 91.3%.

PR #120, opened 2026-06-25
feat: WindRecipe + 5 design-system primitives

anilcancakir:
Adds `WindRecipe`/`WindSlotRecipe` (tailwind-variants `tv()` equivalent, strict emission order, no dedupe/sort/twMerge) and 5 className-driven primitives `WBadge`/`WCard`/`WSwitch`/`WRadio`/`WTabs`, all additive. Full five-surface sync (doc/, example pages, wind-ui skill 22->27 widgets, CHANGELOG, README). analyze 0, 1538 tests, 91.3% coverage.

Cross-repo: this design-first component system spans 5 repos. Land in dependency order: **wind** (WindRecipe + 5 primitives) first, then **magic** (design:sync/lint, previews:refresh, make:component) and **magic_devtools** (preview catalog) which consume wind, then **magic_starter** (component library + view rewrite), then **magic_example** (consumer app). Branches are all `feat/design-first-component-system`. Verified locally end-to-end (analyze 0, full suites green, real-browser e2e of the /preview catalog, release-strip confirmed).

## Summary by CodeRabbit

* **New Features**
  * Added new widgets: `WBadge`, `WCard`, `WSwitch`, `WRadio<T>`, and `WTabs`.
  * Introduced `WindRecipe` / `WindSlotRecipe` for strict, ordered variant and slot-based class composition.
  * Added example pages demonstrating the new widgets and recipe styling.

* **Bug Fixes**
  * Fixed `WSelect` web regression where the dropdown could dismiss on the same click that opened it.

* **Documentation**
  * Added/expanded widget and recipe documentation, and updated README, changelog, and skill materials.

* **Tests**
  * Added comprehensive recipe and widget tests, including interaction, accessibility/semantics, and `WSelect` behavior.


anilcancakir:
Thanks for the review. Pushed `cef63c2` addressing the findings:

**Behavior (Copilot)**
- `WSwitch` / `WRadio` / `WTabs`: `onChanged == null` now means non-interactive **and** disabled: no gesture handler is attached, the `disabled:` className state activates, and Semantics reports `enabled: false`. Matches the documented contract. Added tests for each.
- `WTabs`: added a constructor `assert` for `tabs.isNotEmpty` and `selectedIndex` in range. This drops the `const` constructor (a const assert cannot read `tabs.length`); the only `const WTabs` sites were in this PR's own tests, now updated. No sibling-repo `const WTabs` usage exists.

**Recipe lint (Copilot)**
- The shorthand/longhand debug lint cache now keys on the actual offending token set (sorted), so two distinct offending recipes each warn once instead of one global warning per family pair. Added tests for distinct-offense + idempotency.

**Examples / docs**
- Replaced the no-op `flex-wrap` with Wind's `wrap` utility in the two example pages this PR adds (`wind_recipe_basic.dart`, `w_badge_basic.dart`). Note: several pre-existing example pages (`responsive/`, `backgrounds/`, etc.) also use `flex-wrap`; those predate this PR and are left for a separate cleanup to keep this diff scoped.
- `WCard` dartdoc uses `onTap` (not `onPressed`); `WBadge` dartdoc examples now include `dark:` pairs; `doc/styling/wind-recipe.md` fenced block now has a `text` language (markdownlint MD040).

**CI: `Assert WASM-ready (pana)`**
This step (`deploy.yml`) globally activates the latest `pana` and asserts `is:wasm-ready`. The new code is pure `package:flutter/widgets.dart` + `foundation.dart` (no `dart:io`/`html`/`ffi`/`js`), so it cannot remove wasm-readiness; master passed the same assertion on an earlier pana build, so this looks like pana-version drift rather than a regression from this PR. It is not a required/branch-protected check. Happy to pin pana or dig further if you'd prefer.

Gates after the push: `flutter analyze` 0, `dart format` clean, `flutter test` 1548 pass (1 pre-existing skip), coverage 91.3%, `cd example && flutter analyze` 0.

anilcancakir:
## Review pass: dispositions

Pushed five fix commits (`4d295cb`, `7754b50`, `7aa92c8`, `bea09f3`, `11318c4`, `585d7f7`) addressing the open review threads. CI was red on the **WASM-ready (pana)** gate; that is now fixed too. Full local gate is green: `dart analyze` 0 issues, `dart format` 0 diff, 1551 tests pass, coverage 91.3%.

### CI fix (root cause, not introduced by this PR)
The `Assert WASM-ready (pana)` step ran an unpinned `pana`, which jumped to `0.23.13`. That release bumps its bundled analyzer to 13.x, which mis-evaluates `flutter_svg`'s correct `dart.library.js_interop` conditional export and reports a false-positive `is:wasm-ready` miss. A real `flutter build web --wasm` of the example compiles cleanly, so the package is genuinely WASM-compatible (master is affected identically). Pinned pana to `0.23.12` with an explanatory comment in `deploy.yml`; an upstream pana issue is the proper long-term fix.

### Fixed in this pass
- **WSelect custom triggers**: wrapped `triggerBuilder` / `multiTriggerBuilder` in the shared `TapRegion` group so a tap on a custom trigger no longer self-closes the menu.
- **WRadio centering**: added `flex` to the ring-shell className (default + docs + examples). Without `flex`, a single-child `WDiv` takes the block pass-through path and the `items-center justify-center` tokens were silently ignored.
- **CHANGELOG**: restored the `## [1.1.2] - 2026-06-25` header (a merge had left the released WPopover fix under `[Unreleased]` as a duplicate `### Fixed`).
- **SKILL.md**: the compound-variant example now declares `lg` so it is reachable; corrected the WSwitch thumb-move contract (`justify-start` -> `checked:justify-end`, not `translate-x-*`, which Wind has no parser for); paired `bg-blue-600` with its dark variant in the form/empty-state button examples.
- **Docs/examples**: fixed the reversed shorthand/longhand caption in `wind-recipe.md`; made the `w_radio_basic` plan row forward taps to the same state change; added dark thumb colors in `w_switch_basic`.

### Already addressed in earlier commits of this PR (now outdated)
The null-`onChanged` non-interactive handling (WSwitch / WRadio / WTabs), the `WTabs` range assert, the recipe shorthand/longhand lint cache key, `WCard` dartdoc `onTap`, `WBadge` dartdoc dark pairs, `flex-wrap` -> `wrap`, and the fenced-block language were all fixed in `cef63c2` / `50c3bed` / `dc14a03` before this pass. The corresponding threads are stale.

### Intentional / not changed (rationale)
- **Props-table `Required` column**: the repo doc convention (`.claude/rules/docs.md`) puts `**Required**` in the `Default` column, exactly as these docs do. A separate column would diverge from the standard.
- **Recipe value validation**: values are intentionally permissive (tailwind-variants semantics): an undeclared value emits no class, only an unknown *axis* throws. This is documented in `_resolveSelection` and backed by tests; validating values would break compound-condition matching against arbitrary resolved values.
- **27-widget count / `WFormMultiSelect`**: `WFormMultiSelect` is a public class in `w_form_select.dart`, which the barrel exports, so all 27 widgets are exposed.
- **`WTabs` non-const constructor**: verified that a `const` constructor with the `selectedIndex < tabs.length` assert fails to const-instantiate (`const_eval_property_access`). The non-const choice and its doc comment are correct.
- **`text-white` / `bg-transparent` in the `_button` recipe**: `text-white` on saturated/colored surfaces follows the repo's own header convention (`example-pages.md` uses bare `text-white`); `bg-transparent` is mode-agnostic.
- **Formatting threads** (`w_tabs_basic`, `w_card_basic`, `wind_recipe_basic`): `dart format --set-exit-if-changed` passes with no diff.

anilcancakir:
@coderabbitai full review

anilcancakir:
## Re-review pass (Copilot)

Addressed the four new threads in `0b4a046` / `bea...`:

- **`w_tabs.dart` null `onChanged`**: fixed. A null `onChanged` now marks every tab disabled (`isDisabled` + the `disabled` state), so the tab reports `enabled: false`, suppresses hover/focus, and activates `disabled:` tokens via the state provider, matching `WSwitch` / `WRadio`.
- **`w_switch.dart` thumb/track className**: fixed. `thumbClassName` now passes through as `null` instead of `''` (no needless empty-string cache key); the track className no longer interpolates a trailing space when the caller className is `null`.
- **`w_tabs_basic.dart` / `w_card_basic.dart` formatting**: not actionable. `dart format --output=none --set-exit-if-changed` reports 0 changed on both files, and the CI "Verify formatting" step passes. The files are already format-clean.

Gate: `dart analyze` 0 issues, `dart format` 0 diff, 1551 tests pass, coverage 91.4%.

anilcancakir:
@coderabbitai full review

anilcancakir:
## Re-review pass 2 (Copilot)

All 13 new threads were valid (consistency with the package's own conventions). Addressed in `526bfd8` and the two preceding commits:

- **`onChanged` optional** (`WSwitch`, `WRadio`): dropped `required` so a null/omitted callback expresses a non-interactive control, matching `WCheckbox` / `WInput` / `WSelect`. Docs (Props + Constructor) updated to match.
- **WSwitch debug className**: the effective track className (`flex items-center` + caller className) is now built once and parsed/logged, so debug output matches the applied classes.
- **Dark pairs in docs**: paired the state-prefixed color tokens (`hover:` / `selected:` / `focus:`) and the switch thumb `bg-white` with dark variants across `w-switch.md`, `w-radio.md`, and `w-card.md`. The example pages already followed this convention; the docs now match.

Gate: `dart analyze` 0 issues, `dart format` 0 diff, 1551 tests pass, coverage 91.3%.

anilcancakir:
@coderabbitai full review

anilcancakir:
## Re-review pass 3 (Copilot): all 8 are false positives

All eight new threads claim the example files are not `dart format`-stable and will fail CI. They are verified incorrect and no code change is warranted:

- `dart format --output=none --set-exit-if-changed` on `w_tabs_basic.dart`, `wind_recipe_basic.dart`, `w_card_basic.dart`, `w_radio_basic.dart` reports **0 changed** (exit 0).
- The `dart format` PostToolUse hook reformats every `.dart` edit in this repo, so committed files are always formatter-output.
- The latest CI run on this branch (`28235383866`) is **green**, including the "Verify formatting" step that runs the same `dart format --set-exit-if-changed` gate.

Copilot's code-review does not execute the formatter; these trailing-comma / `createState` wrapping claims are heuristic guesses that disagree with the actual tool output. Resolving as not-actionable. If `dart format` ever does rewrite these files in CI, that is the signal to act, not the review heuristic.

PR #130, opened 2026-06-30
feat(cursor): add cursor-* utilities (#125)

anilcancakir:
## Summary

Adds Tailwind-style `cursor-*` utilities to Wind. Resolves #125.

A new `CursorParser` maps cursor names to `SystemMouseCursors`, `WindStyle` gains a `mouseCursor` field, and `WDiv` wraps its content in a `MouseRegion` when a `cursor-*` token is present. A plain `WDiv` can now show a pointer cursor on web/desktop without `WAnchor` or a manual `MouseRegion`.

### Behavior
- Full Tailwind cursor set mapped to native cursors (pointer→click, not-allowed→forbidden, grab/grabbing, zoom-in/out, the directional + corner resize family, etc.). `cursor-auto`/`cursor-default`→basic (Flutter has no "auto").
- On a `hover:`/`focus:`/`active:` `WDiv` (auto-wrapped in `WAnchor`, whose default cursor is click), the cursor `MouseRegion` sits deeper in the tree and wins.
- Inert on touch platforms; last token wins; unknown `cursor-*` names are a silent no-op.

### Tests
- `test/parser/parsers/cursor_parser_test.dart`: mapping, last-wins, unknown no-op, and WDiv integration (plain + interactive + cursor-less).

### Definition of done
- `dart analyze`: clean. `dart format`: no diff.
- `flutter test`: 1561 pass (1 pre-existing skip), coverage 91.3% ≥ 90%.
- Post-change sync (5 surfaces): `doc/interactivity/cursor.md`, example `cursor_basic` page + route, skill (`SKILL.md` §4 + `tokens.md` §16 + divergence notes, version 2.8.0), `CHANGELOG.md`, `README.md` parser count 19→20.

## Summary by CodeRabbit

* **New Features**
  * Added support for `cursor-*` styles on web and desktop, including common pointer, drag, zoom, and resize cursors.
  * Added a new example page and documentation showing how to use cursor styling, precedence rules, and platform behavior.

* **Bug Fixes**
  * Improved cursor handling so the last matching cursor class wins.
  * Unknown cursor values now safely do nothing, and cursor styling remains inactive on touch devices.

* **Documentation**
  * Updated the changelog, token references, migration guidance, and release notes to reflect cursor support.

PR #131, opened 2026-06-30
fix(alias): expand state/breakpoint-prefixed alias tokens (#124)

### LINKED ISSUE #124: State/breakpoint-prefixed semantic aliases silently no-op (`hover:bg-<alias>`, `sm:bg-<alias>`)
## Problem

A semantic alias key combined with a state or breakpoint prefix (`hover:bg-surface-container`, `active:bg-surface-container`, `sm:bg-surface-container`) silently does nothing — no expansion, no warning. Plain `bg-surface-container` works; `hover:` + a raw token works; but `prefix:` + ALIAS is dropped.

## Current behavior

The alias expander rejects any token containing `:` before lookup (`lib/src/parser/alias_expander.dart:96-100`: `if (token.contains(':')) { out.add(token); return; }`), and expansion runs before prefix resolution (`lib/src/parser/wind_parser.dart:166-172`). So a prefixed alias never reaches alias expansion, and the prefixed-raw form has no matching utility → silent no-op.

## Proposed direction

Expand the alias portion of a prefixed token: split off the prefix (`hover:` / `sm:` / `active:` / …), expand the alias, then re-apply the prefix to each produced utility. Alternatively (lower effort), emit a dev-mode warning when a prefixed token's body is a known alias key that was skipped.

## Workaround

Apply the prefix in Dart with conditional classNames (`isHovering ? 'bg-surface-container' : ''`) instead of `hover:bg-surface-container`.

---
_Surfaced repeatedly in uptizm (sidebar active-nav highlight, hover rows). The silent no-op is the worst part: the code looks correct but does nothing at runtime._

anilcancakir:
## Summary

Fixes #124. A semantic alias combined with a state or breakpoint prefix (`hover:bg-surface-container`, `md:row`, `dark:hover:row`) silently did nothing: the alias expander rejected any token containing `:` before the lookup, so the prefixed-alias form never expanded and the prefixed-raw form had no matching utility.

## Fix

`expandAliases` now peels the prefix chain (everything up to and including the last `:`) off each token, matches the **bare body** against the alias map (keys stay bare-only), and on a hit re-applies the prefix to every produced token:

- `md:row` → `md:flex md:flex-row`
- `hover:bg-surface` → each token the `bg-surface` alias produces carries `hover:`, including the value's own `dark:` peer (`hover:dark:...`), which the parser resolves regardless of prefix order.

Cycle / depth / output-budget guards are unchanged and now key on the bare body (a cycle reached through a prefixed token is still caught).

## Tests

`test/parser/alias_expander_test.dart`: added prefixed expansion, stacked-prefix, prefixed-token cycle, and non-alias-prefixed-passthrough cases. Updated the one prior assertion that encoded the old (buggy) `md:row` no-op behavior.

## Definition of done

- `flutter test`: 1555 pass (1 pre-existing skip), coverage 91.4% ≥ 90%.
- Post-change sync: `doc/core-concepts/theming.md`, skill (`SKILL.md` alias note, version 2.8.0), `CHANGELOG.md`. README not touched (bug fix, not overview-worthy).

> Note: the `SKILL.md` version bump to 2.8.0 overlaps with #130 (cursor) which also bumps from 2.7.0→2.8.0. Whichever merges second will have a one-line version conflict on that frontmatter line; trivially resolved (keep 2.8.0).


anilcancakir:
Fixed in the latest commit: the residual `[!NOTE]` callout (it still said prefix variants are not expanded) now describes the peel-prefix / match-bare-body / re-apply-prefix behavior, consistent with the bullet above it.

@coderabbitai review

PR #132, opened 2026-06-30
feat(sizing): add size-* utility, fix childless WDiv sizing (#123)

### LINKED ISSUE #123: Childless `WDiv` ignores `size-*`/`w-*`/`h-*` and collapses to 0×0
## Problem

A `WDiv` with explicit sizing but no `child` / `children` renders at 0×0, ignoring the size utilities. This breaks common cases like a status dot or a fixed-size spacer.

## Repro

```dart
WDiv(className: 'size-2 rounded-full bg-up'); // renders 0x0, invisible
```

## Current behavior

With both `child` and `children` null, `_buildCoreStructure` returns `null` (`lib/src/widgets/w_div.dart:457-459`); the composition pipeline ends at `widgetToBuild ?? const SizedBox.shrink()` (`lib/src/widgets/w_div.dart:1602`), a 0×0 box. Explicit `size-*` / `w-*` / `h-*` is not applied to an empty wrapper.

## Proposed direction

When a WDiv has explicit width/height (including `size-*`) but no children, honor that sizing (e.g. `SizedBox(width, height, child: <coloredBox>)`), matching a CSS empty `<div>` with width/height.

## Workaround

Wrap in `SizedBox(width:.., height:..)`, or pass `child: const SizedBox.shrink()`.

---
_Surfaced in a StatusBadge: the leading dot (`size-1.5 rounded-full bg-up`) was invisible until wrapped in a `SizedBox`._

anilcancakir:
## Summary

Fixes #123. A childless `WDiv(className: 'size-2 rounded-full bg-...')` (a status dot) rendered wrong / collapsed.

### Root cause
The real gap was that **`size-*` was never a recognized token**, not that `w-*`/`h-*` were ignored. Verified by reproduction: a childless `WDiv(className: 'w-10 h-10 ...')` already renders 40×40 correctly (the composition pipeline applies the `SizedBox`/`Container` whether or not there is content); `size-2` was simply dropped as unknown, so the box took its decoration with no size and collapsed in a flex context.

### Fix
Adds the Tailwind v3.4+ `size-*` shorthand to `SizingParser`. `size-N` (theme spacing), `size-full`, `size-1/2` (fraction), `size-[20px]` / `size-[50%]` (arbitrary), and `size-screen` set **both** width and height in one token. A later `w-*`/`h-*` overrides the matching axis (reverse-iteration last-wins).

### Tests
- `sizing_parser_test.dart`: size-N / size-full / fraction / arbitrary / screen / canParse / w-* override.
- `w_div/sizing_test.dart`: childless `size-*` renders 8×8 in Center and inside a flex row; childless `w-*/h-*` confirmed working.

### Definition of done
- `dart analyze`: clean (lib + test + example page). `dart format`: no diff.
- `flutter test`: 1561 pass (1 pre-existing skip), coverage 91.4% ≥ 90%.
- Post-change sync: `doc/layout/sizing.md` (Size section), example `sizing_basic` demo, skill (`SKILL.md` + `tokens.md` §4, version 2.8.0), `CHANGELOG.md`. README not touched (token within an existing parser family).

> Note: `SKILL.md` version bump to 2.8.0 overlaps with #130/#131; whichever merges last takes a trivial one-line version conflict (keep 2.8.0).


anilcancakir:
Thanks for the careful review. Addressed and one deliberate decline:

**Fixed:** added the implemented `size-screen` form to the `tokens.md` size row and clarified the last-wins wording.

**Declined (with rationale) — the four 'guard absolute vs factor per axis' suggestions:** Wind's sizing parser intentionally allows an absolute value and a fractional value to coexist on the same axis; this is an established contract, not a `size-*` regression. The existing test `SizingParser > parses width classes correctly` asserts that `['w-100', 'w-[50%]']` yields BOTH `width: 400` AND `widthFactor: 0.5`. `size-*` deliberately mirrors `w-*`/`h-*` here: `size-full w-10` behaves exactly like `w-full w-10`. Adding a single-mode guard only for `size-*` would make it stricter than `w-*`/`h-*` and break that test. Same-axis same-mode last-wins (the realistic case, `size-2 w-8` -> width 32) already works and is covered by `a later w-* overrides the width set by size-*`. Re-tightening the absolute-vs-factor model for ALL sizing tokens is a separate, broader change I'd rather not fold into this fix.

@coderabbitai review

PR #133, opened 2026-06-30
fix(popover): bound overlay width with width/maxWidth props (#127)

### LINKED ISSUE #127: `WPopover` overlay has no `maxWidth`/`width` — menus stretch unbounded (asymmetry with `WSelect`)
## Problem

`WPopover`'s overlay applies only `minWidth` (trigger width) and no `maxWidth`, so a popover/menu without an explicit width stretches to fill available space (observed ~900px on an 800px content column, overflowing). `WSelect` does not have this problem because it sets an explicit menu width.

## Current behavior

`WPopover` overlay: `ConstrainedBox(constraints: BoxConstraints(maxHeight: ..., minWidth: parsedWidth ?? triggerWidth))` — no `maxWidth` (`lib/src/widgets/w_popover.dart:657-662`).
`WSelect` resolves `menuWidth ?? menuStyles.width ?? triggerWidth` and applies it as a fixed `width` (`lib/src/widgets/w_select.dart:722-758`).

## Proposed direction

Give `WPopover` a `maxWidth` (and/or `width`) option, or default the overlay to a bounded width like `WSelect`, so dropdown menus built on `WPopover` don't overflow without a manual `w-*`.

## Workaround

Pass an explicit `w-*` (e.g. `w-64`) on the popover panel / className.

---
_Surfaced with a "Last 24 hours" date-range dropdown in uptizm (built on a `DropdownMenu` that wraps `WPopover`) which stretched full width, vs the form picker on `WSelect` that rendered correctly. This `WPopover` ↔ `WSelect` asymmetry is the root cause._

anilcancakir:
## Summary

Fixes #127. `WPopover`'s overlay applied only a `minWidth` (trigger width) and no `maxWidth`, so a menu without an explicit width stretched to fill available space and could overflow a narrow content column (observed ~900px on an 800px column). `WSelect` avoids this by pinning a menu width; this restores parity.

## Fix

Two new props:
- `width` (`double?`) — pins a fixed overlay width, like `WSelect`'s `menuWidth`. Overrides a `w-*` token in `className`.
- `maxWidth` (`double?`) — upper bound. Falls back to a `max-w-*` token in `className`, then to the screen width, so the overlay never overflows the viewport.

Resolution: a fixed `width` / `w-*` pins `min == max`; otherwise the overlay sizes between the trigger width (clamped to the bound) and `effectiveMaxWidth`.

## Tests

`test/widgets/w_popover/width_test.dart`: default maxWidth = screen width, explicit `maxWidth`, fixed `width`, `w-*` token, `max-w-*` token, and `width` overriding `max-w-*`. (The overlay ConstrainedBox is selected by its `maxHeight` to distinguish it from the content WDiv's own `max-w-*` box.)

## Definition of done

- `dart analyze`: clean (lib + test + example). `dart format`: no diff.
- `flutter test`: 1557 pass (1 pre-existing skip), coverage 91.4% ≥ 90%.
- Post-change sync: `doc/widgets/w-popover.md` (props + constructor), example `popover_basic` maxWidth demo, skill (`references/widgets.md` + `SKILL.md` version 2.8.0), `CHANGELOG.md`. README not touched (props on an existing widget).

> Note: `SKILL.md` 2.8.0 bump overlaps with #130/#131/#132; last to merge takes a trivial one-line version conflict (keep 2.8.0).


anilcancakir:
All four addressed in the latest commit:
- Clamped auto-flip `popoverWidth` to `effectiveMaxWidth` (else `fixedWidth`), so the sizing fed to `computeEffectiveAlignment` matches the actual overlay constraint and no longer over-flips when `triggerWidth > maxWidth`. Removed the redundant `parsedWidth` alias.
- Fixed the `width` dartdoc: dropped the broken `[minWidth]` link, describe the trigger-width minimum in prose.
- Renamed the `_wrap` test helper to the repo-standard `wrapWithTheme` (both call sites).

@coderabbitai review

anilcancakir:
Follow-up: fixed the codecov/patch failure my previous commit introduced. The auto-flip clamp had landed in the `_buildOverlay` fallback (only reached when `_effectiveAlignment` is null at build, which the pre-calc normally prevents), so those lines were uncovered. Moved the clamp into `_calculateEffectiveAlignment` — the path that actually computes the alignment on open — which is exercised by the width tests, and added a wide-trigger + small-maxWidth test. Patch should be green now.

anilcancakir:
Addressed the two follow-up comments:
- **Negative/NaN/Infinity width guard:** `width`/`maxWidth` are now sanitized (finite & > 0) before reaching `BoxConstraints`; an invalid value is dropped and the overlay falls back to flexible sizing. Added tests for negative `width` and infinite `maxWidth`.
- **Dark peer:** added `dark:shadow-xl` to the maxWidth example's className.

Also merged latest master (CHANGELOG conflict resolved, keeping both entries).

@coderabbitai review

PR #134, opened 2026-06-30
feat(tabs): default WTabs list to full container width (#128)

### LINKED ISSUE #128: `WTabs` tab list is content-width — underline/`border-b` not full-width by default
## Problem

`WTabs` renders the tab list as a content-sized row, so a `border-b` underline on the list spans only the tabs' width, not the full container. A full-width baseline (the common tab design) requires the caller to add `w-full` to `listClassName`.

## Current behavior

The tab list is `WDiv(className: listClassName, children: [...])` (`lib/src/widgets/w_tabs.dart:133-138`); a `flex-row` `WDiv` defaults to `MainAxisSize.min`, so without `w-full` it is content-width.

## Proposed direction

Default the tab list to full container width, or expose a `fullWidthList` / `underline` option, so the baseline rule spans the container by default. At minimum, document the `w-full` requirement on `listClassName`.

## Workaround

Add `w-full` to `listClassName`.

---
_Surfaced in uptizm's monitor-detail tab bar (Overview / Metrics / Incidents): the underline stopped short of the content width._

anilcancakir:
## Summary

Fixes #128. `WTabs` rendered the tab list as a content-width `flex-row` `WDiv`, so a `border-b` underline on `listClassName` spanned only the tabs' width, not the container. The full-width baseline (the common tab design) required the caller to add `w-full` to `listClassName` by hand.

## Fix

New `fullWidthList` prop (default `true`) prepends `w-full` to the tab list so the underline spans the full container width. The tabs themselves stay content-sized and left-aligned (default `justify-start`). Set `fullWidthList: false` for a content-width strip (centered / pill tabs) or when driving the list width with an explicit `w-*` token.

`WTabs` is still unreleased (`[Unreleased]` / Added), so changing the default is not a breaking change for any published release.

## Tests

`test/widgets/w_tabs_test.dart` (new group): list spans full width by default, `fullWidthList: false` stays content-width, and `fullWidthList: false` + `w-32` honors the explicit width.

## Definition of done

- `dart analyze`: clean (lib + test + example). `dart format`: no diff.
- `flutter test`: 1554 pass (1 pre-existing skip), coverage 91.4% ≥ 90%.
- Post-change sync: `doc/widgets/w-tabs.md` (prop + constructor), example `w_tabs_basic` pill demo opts out via `fullWidthList: false`, skill (`SKILL.md` WTabs row + version 2.8.0), `CHANGELOG.md` WTabs entry. README not touched (WTabs already listed).

> Note: `SKILL.md` 2.8.0 bump overlaps with the sibling PRs; last to merge takes a trivial one-line version conflict (keep 2.8.0).


anilcancakir:
Good catch — rather than only document it, I made it deterministic.

`w-full` sets `widthFactor` and an explicit `w-*` sets `width` (different `WindStyle` fields), so prepending both really is ambiguous (the pipeline applies both, fractional + SizedBox). WTabs now **skips** the `w-full` prepend when `listClassName` already carries a width token (new `_hasWidthToken` helper), so an explicit width wins cleanly, and the default still spans full width when no width is set. Updated the dartdoc + doc table and added a test asserting `w-32` wins with `fullWidthList: true`.

@coderabbitai review

PR #136, opened 2026-06-30
docs(layout): document the intrinsic-sizing limitation (#129)

### LINKED ISSUE #129: Document that `IntrinsicHeight`/intrinsic sizing fails through Wind content (LayoutBuilder)
## Problem

Wrapping Wind widgets in `IntrinsicHeight` (or a `Row` / `Column` that needs child intrinsic heights, e.g. for equal-height columns) throws "LayoutBuilder does not support returning intrinsic dimensions". This is not discoverable until you hit it at runtime.

## Current behavior

Wind wraps content in a `LayoutBuilder` for `h-full` sizing (`lib/src/widgets/w_div.dart:1431-1455`) and for flex `basis-*` resolution (`lib/src/widgets/w_div.dart:531-554`). `LayoutBuilder` runs during the layout phase, not the intrinsic-sizing phase, so intrinsic queries through Wind content assert. (This is a fundamental Flutter constraint, not a wind bug.)

## Proposed direction

Document this limitation in the layout/sizing docs, with the recommended escape hatches (explicit sizing; avoid `IntrinsicHeight` / `IntrinsicWidth` around Wind content). Optionally, offer an opt-in non-`LayoutBuilder` path for cases that genuinely need intrinsic sizing (relates to the grid equal-height enhancement).

## Workaround

Use explicit sizing (`w-*` / `h-*`) instead of intrinsic queries; do not wrap Wind content in `IntrinsicHeight` / `IntrinsicWidth`.

---
_Surfaced trying to equalize KPI card heights in uptizm: the `IntrinsicHeight` approach threw, forcing a reserve-placeholder workaround instead._

anilcancakir:
## Summary

Fixes #129. Wrapping Wind content in `IntrinsicHeight` / `IntrinsicWidth` (or a `Row`/`Column` that needs child intrinsic heights for equal-height columns) throws `LayoutBuilder does not support returning intrinsic dimensions`, and this was only discoverable at runtime.

## Root cause (now documented)

To resolve `h-full` and flex `basis-*` against the parent's real constraints, Wind wraps that content in an internal `LayoutBuilder`. `LayoutBuilder` runs in the layout phase, not the intrinsic-sizing phase, so any intrinsic query passing through such content asserts. This is a fundamental Flutter constraint, not a Wind bug. Content that does not resolve `h-full` / `basis-*` carries no `LayoutBuilder` and is safe.

## Changes

- `doc/layout/sizing.md`: new **Intrinsic Sizing Limitation** section (symptom, cause, what triggers it, escape hatches — explicit `h-*`/`size-*`, `Stack` + `Positioned(top:0,bottom:0)` for equal-height rows, and the intrinsic-free `items-stretch` column) + ToC entry.
- Skill: `SKILL.md` §6 limitation note + `references/layouts.md` items-stretch caveat, version 2.8.0.

## Definition of done

- Doc-only: no `lib/` change, so `CHANGELOG.md`, tests, and coverage are intentionally untouched.
- `dart format --set-exit-if-changed`: clean (0 changed). `dart analyze`: clean. No new `<x-preview>` added (section uses inline code), so no example page needed.

> Note: `SKILL.md` 2.8.0 bump overlaps with the sibling PRs; last to merge takes a trivial one-line version conflict (keep 2.8.0).


anilcancakir:
Fixed all three: reworded `sizing.md`, `SKILL.md`, and `layouts.md` to match the implementation precisely — `h-full` adds a `LayoutBuilder` only in an unbounded-height context, and `basis-*` adds a single `LayoutBuilder` around the surrounding flex (not the individual child).

@coderabbitai review

anilcancakir:
Addressed the precision concerns + merged master:
- Removed the "intrinsic-free and safe" overclaim. Wind's column cross-axis stretch DOES use a `LayoutBuilder` internally; it's a REPLACEMENT for `IntrinsicHeight` (you don't wrap it in one), not something safe to nest inside an `IntrinsicHeight`. Reworded in `SKILL.md` and `sizing.md`.
- The "safe to wrap" sentence now lists the column-stretch `LayoutBuilder` as another path, so it no longer implies only `h-full`/`basis-*` are affected.

@coderabbitai review

Issue #126, opened 2026-06-30
`grid` does not stretch cells to equal height (no `items-stretch` for grid)


anilcancakir:
## Problem

Wind's `grid` sizes each cell to its own content; cells in the same row are NOT stretched to the row's tallest (CSS Grid's default `align-items: stretch`). A row of cards where one has an extra line ends up ragged.

## Current behavior

Grid is built with `Wrap` + a per-item `SizedBox` constraining width only (`lib/src/widgets/w_div.dart:1032-1086`; the comment at ~`:1028` notes the deliberate Wrap-based intrinsic-height choice). Height is never equalized across a row, and `h-full` on a cell has no bounded parent height to resolve against.

## Proposed direction

Support equal-height grid rows: an `items-stretch`-equivalent, or a grid layout mode that stretches cells in a row to the max height. Relatedly, make `h-full` resolvable inside a grid cell. (Related to the `IntrinsicHeight` limitation documentation issue.)

## Workaround

Reserve equal content in each cell (e.g. always render placeholder footer rows) so intrinsic heights match.

---
_Surfaced with a KPI stat-card row in uptizm: one card had a delta row and rendered taller than the rest._

PR #140, opened 2026-07-01
fix(grid): stretch unequal flex-col cells without the IntrinsicHeight assert (#139)

### LINKED ISSUE #139: `items-stretch` asserts (RenderIntrinsicHeight) when stretching an unequal-height `flex flex-col` cell
## Problem

`grid ... items-stretch` (added in #126 / #137) asserts with a `RenderBox was not laid out` cascade (rooted at `RenderIntrinsicHeight`) when it has to STRETCH an unequal-height cell whose content is a `flex flex-col` (for example a card built as a column). Since stretching unequal cells is the whole point of `items-stretch`, this bites exactly the case the feature exists for.

## Trigger

- A `grid ... items-stretch` whose cells have DIFFERENT natural heights (so `IntrinsicHeight` must stretch the shorter ones).
- Each cell's content is a `flex flex-col` WDiv (a common card shape: label + value + optional rows).

When all cells happen to be equal height, `IntrinsicHeight` never stretches and it renders fine — so the bug hides until a row has a taller sibling.

## Why

`items-stretch` builds the grid as a column of `IntrinsicHeight` rows (#126). When `IntrinsicHeight` stretches a shorter cell, the stretched `flex flex-col` cell content inserts a `LayoutBuilder` (Wind uses a `LayoutBuilder` for flex/height resolution — same construct behind the #129 intrinsic-sizing limitation). `LayoutBuilder` cannot report intrinsic dimensions, so the dry-layout/intrinsic pass asserts (`box.dart:2251 hasSize` / `object.dart:2841 !_debugDoingThisLayout`).

## Repro (shape)

```dart
WDiv(
  className: 'grid grid-cols-2 gap-4 items-stretch',
  children: [
    // cell A: 2 rows
    WDiv(className: 'flex flex-col gap-1', children: [WText('A'), WText('1')]),
    // cell B: 3 rows (taller) -> forces IntrinsicHeight to stretch cell A -> assert
    WDiv(className: 'flex flex-col gap-1', children: [WText('B'), WText('2'), WText('note')]),
  ],
)
```

Observed on Flutter web (CanvasKit) via a real app (uptizm): a 4-card KPI grid where two cards had an extra caption row crashed the whole screen (blank render, 18-error assertion cascade); the same grid with equal-height cards rendered fine.

## Expected

`items-stretch` should stretch an unequal `flex flex-col` cell to the row's tallest without asserting — that is its purpose. Either the stretched-cell path must avoid the intrinsic-incompatible `LayoutBuilder`, or `items-stretch` should use a stretch mechanism that does not depend on intrinsic sizing of `LayoutBuilder`-bearing content.

## Workaround

Do not use `items-stretch` for equal-height when cells contain `flex flex-col` content and can differ in height; equalize height another way (e.g. reserve/pad the shorter cells to a uniform row count). This is the pre-#126 approach.

## Related

- #126 / #137 (the `items-stretch` feature)
- #129 (`IntrinsicHeight` cannot measure through Wind's `LayoutBuilder`) — same root construct.

---
_Surfaced adopting #126 in uptizm to replace a reserved-footer equal-height hack: the detail KPI row (equal cells) worked, the dashboard KPI row (unequal cells) crashed. Reverted to the reserved-footer hack pending this fix._

anilcancakir:
## Summary

Fixes #139, a regression in the #126/#137 `items-stretch` grid. `grid ... items-stretch` crashed with `RenderBox was not laid out` (rooted at `RenderIntrinsicHeight` → `LayoutBuilder does not support returning intrinsic dimensions`) the moment it had to **stretch an unequal `flex flex-col` cell** — i.e. the exact case the feature exists for. Equal-height cells hid the bug; a taller sibling triggered it.

### Root cause (reproduced first)
The stretch grid wrapped each row in `IntrinsicHeight` + `Row(crossAxisAlignment: stretch)`. `IntrinsicHeight` measures children via the **intrinsic protocol**. A `flex flex-col` cell (or one using `h-full` / `basis-*`) carries a `LayoutBuilder`, and `LayoutBuilder` cannot answer intrinsic (or dry-layout) queries — so stretching a shorter cell asserted. Same root construct as #129.

### Fix
Replace `IntrinsicHeight` + `Row(stretch)` with **`WindEqualHeightRow`**, a small `RenderObject` that:
1. lays each cell out with a **real** loose-height layout (which `LayoutBuilder` fully supports) to find the tallest,
2. re-lays every cell tight to that max height,
3. gives each cell an equal share of the width (matching the fixed column count), with `spacing` between columns.

Real layout instead of intrinsic queries means LayoutBuilder-bearing cells stretch correctly. **Bonus:** `h-full` inside a stretched cell now resolves too (previously documented as unsupported).

### Tests
- `test/widgets/wind_equal_height_row_test.dart`: stretch-to-tallest, equal width split, `spacing` setter (rebuild), a `LayoutBuilder` child laid out without asserting, hit-testing, unbounded-width degradation, empty.
- `test/widgets/w_div/grid_stretch_test.dart`: new #139 case — unequal `flex flex-col` cells with a nested `WDiv` stretch without asserting and equalize. The existing 5 grid tests still pass.

### Definition of done
- `flutter test`: all pass, coverage 94.4% ≥ 90% (`wind_equal_height_row.dart` fully covered).
- Post-change sync: `doc/layout/grid.md` (mechanism + dropped the now-obsolete "no h-full on a stretched cell" caveat), skill (`tokens.md` + `references/layouts.md`, `SKILL.md` version 2.8.1), `CHANGELOG.md`. `WindEqualHeightRow` is an internal render helper (not exported / not a public W-widget), so the README roster is unchanged.

## Summary by CodeRabbit

* **New Features**
  * Added improved equal-height grid rows that size cells based on the tallest item, with consistent spacing and better support for stretched content.
  * Expanded layout behavior so grids can handle more complex children and unbounded width scenarios more reliably.

* **Bug Fixes**
  * Fixed a crash in equal-height grid layouts involving uneven flex-column cells.
  * Resolved cases where stretched cells could fail to fill the full row height correctly.

* **Documentation**
  * Updated grid and layout guidance to reflect the new equal-height row behavior.


anilcancakir:
Addressed all five review comments:
- **spacing validation:** debug asserts (finite, non-negative) on both the widget constructor and the render-object setter.
- **height constraint:** cells now stretch to `maxHeight.clamp(constraints.minHeight, constraints.maxHeight)`, so a parent-imposed tight/min row height stays consistent with the cells (not just the measured tallest).
- **stale dartdoc:** `_buildStretchGrid` doc rewritten to describe `WindEqualHeightRow` (real layout), dropping the obsolete `IntrinsicHeight` + no-`h-full` caveat.
- **vacuous test:** the LayoutBuilder-child test now asserts the cell's actual stretched height (80), proving real layout rather than a never-null check.

@coderabbitai review

PR #142, opened 2026-07-01
fix(grid): stretch cells with a min height to drop the 2px overflow (#141)

### LINKED ISSUE #141: Follow-up to #139: `items-stretch` emits a 2px bottom overflow on the stretched cell
## Problem

Follow-up to #139. `grid ... items-stretch` no longer crashes when stretching an unequal `flex flex-col` cell (great), but the stretch path now emits a non-fatal `RenderFlex overflowed by 2.0 pixels on the bottom.` on the row that actually gets stretched. The crash is gone; a 2px overflow remains.

## Trigger

- A `grid ... items-stretch` whose cells have DIFFERENT natural heights, so `WindEqualHeightRow` must stretch the shorter cell(s) to the row max.
- Cell content is a `flex flex-col` inside a card (WDiv with padding + `gap-*`).

Observed: a 4-card KPI row where two cards have an extra caption row. The row renders equal-height and looks correct, but throws `RenderFlex overflowed by 2.0 pixels on the bottom`. An equal-height row (no cell needs stretching) throws nothing — verified side by side: the detail KPI row (all cells same height) is clean; the dashboard KPI row (3 vs 4 rows) overflows 2px.

## Likely cause

`WindEqualHeightRow` measures each cell with a loose-height pass, takes the row max, then re-lays every cell tight to that max. The tight re-lay of the cell content (`flex flex-col` with `gap-*` + card padding) appears to need ~2px more than the loose-measured max (a measurement/rounding discrepancy between the loose-measure pass and the tight re-lay), so the content overflows the assigned tight height by 2px.

## Expected

Stretching a cell to the row max should not overflow: the re-laid height must be `>=` the cell's own content height under the same constraints. Reconcile the loose-measure vs tight-relay height (round up / use the measured intrinsic consistently) so a stretched cell has no residual overflow.

## Repro (shape)

```dart
WDiv(
  className: 'grid grid-cols-2 gap-4 items-stretch',
  children: [
    WDiv(className: 'flex flex-col gap-1', children: [WText('A'), WText('1'), WText('note')]), // taller
    WDiv(className: 'flex flex-col gap-1', children: [WText('B'), WText('2')]),                 // stretched -> 2px overflow
  ],
)
```

## Related

- #139 (the WindEqualHeightRow stretch fix this follows up)
- #126/#137 (the items-stretch feature)

---
_Surfaced re-adopting #126/#139 in uptizm: after the #139 fix the dashboard KPI grid stopped crashing and renders equal-height, but emits a 2px bottom overflow on the stretched row. Cosmetically invisible, non-fatal, but a residual layout warning._

anilcancakir:
## Summary

Fixes #141, a follow-up to #139. `grid ... items-stretch` no longer crashes (thanks to #139's `WindEqualHeightRow`), but it emitted a residual `RenderFlex overflowed by 2.0 pixels on the bottom` on the row that actually gets stretched.

### Cause
`WindEqualHeightRow` re-laid each cell to a **tight** height equal to the loose-measured row max. A cell whose `flex flex-col` content needs a hair more under the tight re-lay than the loose pass reported (sub-pixel text/flex rounding on real rendering, e.g. CanvasKit) then overflowed by ~2px.

### Fix
Stretch with a **min** height instead of a tight one: each cell is laid out with `minHeight = row target` and an unbounded (or the incoming) max, so it is **never forced below its own content**, and the row takes the tallest resulting height. Overflow is impossible by construction; the common integer-metric case still yields perfectly equal cells. (This also folds the previous separate measure/relay passes into one positioning pass.)

### Tests / reproducibility
The 2px overflow is specific to fractional glyph metrics (CanvasKit) and is **not reproducible in the `flutter_test` binding** (integer test-font metrics), so I could not write a red→green test for the exact pixel. The added guard (`grid_stretch_test.dart` #141 case) asserts **no exception + equal heights** on the repro shape, and the fix is correct **by construction** (a min-height stretch cannot squeeze content below its own height, so `RenderFlex` cannot overflow).

### Definition of done
- `flutter test`: all pass, coverage 94.4% ≥ 90% (`wind_equal_height_row.dart` fully covered; no conditional re-lay branch left uncovered).
- Post-change sync: `doc/layout/grid.md`, skill (`tokens.md`, `SKILL.md` version 2.8.2), `CHANGELOG.md`. `WindEqualHeightRow` stays internal (README unchanged).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved equal-height grid rows so stretched cells stay aligned without triggering small overflow warnings.
  * Fixed cases with uneven content in nested flex layouts to render cleanly while keeping matching row heights.

* **Documentation**
  * Clarified grid and layout docs to describe the updated stretch behavior and its impact on overflow warnings.

* **Tests**
  * Added coverage to verify stretched unequal cells render without overflow exceptions and remain the same height.


anilcancakir:
Both addressed:
- **Equal heights:** added a settle pass — after the min-height measure pass determines the true row height, every cell is re-laid to that height (still a MIN constraint, never tight). So if any cell grows past the target, the shorter siblings are brought up to match; all cells share the row height and none is squeezed below its content (still overflow-free).
- **Dup comment:** fixed the duplicated `intrinsic query … intrinsic query.` fragment in the `_buildStretchGrid` dartdoc.

@coderabbitai review

PR #156, opened 2026-07-20
fix(theme): tolerate unknown spacing tokens instead of throwing

### LINKED ISSUE #155: Unknown theme-spacing tokens like p-primary throw ArgumentError inside build()
### What happened

`WindThemeData.getSpacing` (`lib/src/theme/wind_theme_data.dart:305-321`) throws `ArgumentError` for any `multiplier` that is not `'full'`, a container key, or a numeric string:

```dart
} else {
  throw ArgumentError('Invalid spacing multiplier: $multiplier');
}
```

The padding, margin, position, and flex-gap parsers each pass the raw regex-captured value directly to `getSpacing` with no pre-validation:

- `lib/src/parser/parsers/padding_parser.dart:86`
- `lib/src/parser/parsers/margin_parser.dart:100`
- `lib/src/parser/parsers/position_parser.dart:100`
- `lib/src/parser/parsers/flexbox_grid_parser.dart:326` (gap-*)

Their regexes admit letters (`[a-zA-Z0-9./]+`), so tokens like `p-primary`, `m-foo`, `mx-red`, `top-abc`, `gap-x-blue` pass `canParse` and match the theme regex, then throw at `getSpacing`. Because the throw happens inside `build()`, release users see a red screen.

The sibling `sizing_parser.dart:207` pre-validates the same regex-captured value with `double.tryParse` before calling `getSpacing`, showing the intended contract is 'unknown tokens are dropped, not fatal.' `WindParser.findAndGroupClasses` also warns-once for unknown tokens, on the same premise.

### Expected

An unknown spacing token is silently dropped (with the existing one-time debug warning that unknown className tokens get), never surfaced as a runtime crash inside `build()`.

### Repro

```dart
WDiv(className: 'p-primary', child: Text('hi'))
// -> throws ArgumentError: Invalid spacing multiplier: primary
```

A user who thinks Wind treats color/theme names as spacing (a natural typo after using `bg-primary`, `text-primary`) crashes their app instead of getting the promised 'unknown className was ignored' warning.

### Fix approach

Add a non-throwing `WindThemeData.tryGetSpacing` sibling and switch the four throwing callers to use it, dropping on `null` via their existing null-guards. `getSpacing` stays unchanged for backward compatibility.

thejesh23:
Fixes #155

## Summary

`WindThemeData.getSpacing` throws `ArgumentError` for any `multiplier` that isn't `'full'`, a container key, or numeric. The padding, margin, position, and flex-gap parsers each pass the raw regex-captured value directly, so tokens like `p-primary`, `m-foo`, `top-abc`, `gap-x-blue` — natural typos after using `bg-primary`/`text-primary` — throw at build time and produce a red screen. This violates the parser contract at `wind_parser.dart:369` (unknown className is dropped with a debug warning); the sibling `sizing_parser.dart` already pre-validates with `double.tryParse`.

Adds `WindThemeData.tryGetSpacing(String)` — same behaviour as `getSpacing` but returns `null` for unrecognised tokens. Switches the four throwing callers to use it and drop the token via their existing `if (value == null) continue;` guards. `getSpacing` is unchanged.

## Definition of done

- [ ] `dart analyze` — please run in CI (no toolchain on audit machine).
- [ ] `dart format .` — please run in CI.
- [ ] `flutter test` — added coverage:
      - `test/theme/wind_theme_data_test.dart` — `tryGetSpacing` matches `getSpacing` for known tokens; returns `null` for unknown.
      - `test/parser/parsers/padding_parser_test.dart` — `p-primary`/`mx-foo`/`py-bar` no longer throw.
      - `test/parser/parsers/margin_parser_test.dart` — `m-primary`/`mx-foo`/`my-bar` no longer throw.
      - `test/parser/parsers/position_parser_test.dart` — `top-abc`/`left-foo`/`inset-x-bar` no longer throw.
      - `test/parser/parsers/flexbox_grid_parser_test.dart` — `gap-blue`/`gap-x-primary`/`gap-y-foo` no longer throw.
- [ ] `./tool/coverage.sh 90` — new `tryGetSpacing` is fully exercised and each patched parser gains a dedicated test, so coverage should hold at ≥ 90 %. Please confirm in CI.

## Post-change sync

This is a pure defect fix; no new widget/parser/token/theme field is added. `doc/`, `example/lib/pages/`, `skills/wind-ui/`, and `README.md` are unchanged per the guidance. CHANGELOG updated under `[Unreleased] > Fixed`.

## Notes

I could not run `flutter test`/`dart analyze`/`dart format` locally (no Flutter toolchain on the audit machine). Please rely on CI. All edits were made against HEAD `aa07198` and follow the parser conventions in `.github/instructions/parsers.instructions.md`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Unknown theme spacing tokens in margin, padding, position, and gap utilities are now safely ignored instead of causing errors.
  * Valid spacing tokens continue to resolve as before, including numeric values and `full`.

* **Tests**
  * Added coverage for unknown spacing tokens across supported utility parsers and theme spacing lookups.

* **Documentation**
  * Updated the changelog with the corrected behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ This project follows [Semantic Versioning 2.0.0](https://semver.org/spec/v2.0.0.
 
 ### Fixed
 
+- Unknown theme-spacing tokens (`p-primary`, `m-foo`, `top-abc`, `gap-x-blue`, ...) now silently drop instead of throwing `ArgumentError: Invalid spacing multiplier: <token>` inside `build()`, matching the "unknown className is dropped with a debug warning" contract other parsers already follow. Adds `WindThemeData.tryGetSpacing(String)`; the padding, margin, position, and flex-gap parsers now use it (sizing already pre-validated with `double.tryParse`). `getSpacing` is unchanged for backward compatibility. (`lib/src/theme/wind_theme_data.dart`, `lib/src/parser/parsers/padding_parser.dart`, `lib/src/parser/parsers/margin_parser.dart`, `lib/src/parser/parsers/position_parser.dart`, `lib/src/parser/parsers/flexbox_grid_parser.dart`)
 - `WAnchor` now hit-tests its whole bounds via `HitTestBehavior.translucent`. The inner `GestureDetector` used the default `HitTestBehavior.deferToChild`, so it only fired when a painted child sat under the exact tap point. An anchor wrapping transparent content (a settings row, a checkbox row, a link with padding) ignored taps that landed on its empty regions, including the element centre that automated drivers and centred pointer events target. It now behaves like `WInput`/`WPopover`, which already use whole-box hit-testing, so the full anchor rectangle is tappable. (`lib/src/widgets/w_anchor.dart`; covered by `test/widgets/w_anchor_test.dart`)
 
 ## [1.2.0] - 2026-07-08
--- a/lib/src/parser/parsers/flexbox_grid_parser.dart
+++ b/lib/src/parser/parsers/flexbox_grid_parser.dart
@@ -323,7 +323,9 @@ class FlexboxGridParser implements WindParserInterface {
           final match = _themeGapRegex.firstMatch(className);
           if (match != null) {
             root = match.namedGroup('root')!;
-            value = theme.getSpacing(match.namedGroup('value')!);
+            // Use `tryGetSpacing` so an unknown token (e.g. `gap-x-blue`)
+            // is dropped by the null-guard below rather than throwing.
+            value = theme.tryGetSpacing(match.namedGroup('value')!);
           }
         }
 
--- a/lib/src/parser/parsers/margin_parser.dart
+++ b/lib/src/parser/parsers/margin_parser.dart
@@ -96,8 +96,9 @@ class MarginParser implements WindParserInterface {
         // Skip 'auto' value - it's handled by mx-auto
         if (valueKey == 'auto') continue;
 
-        // Call the theme's `getSpacing` method
-        value = theme.getSpacing(valueKey);
+        // Use `tryGetSpacing` so an unknown token (e.g. `m-foo`) is dropped
+        // by the null-guard below rather than throwing at build.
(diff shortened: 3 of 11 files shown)


anilcancakir (approved):
Thanks for the thorough writeup, @thejesh23. The diagnosis is exactly right: `getSpacing` throwing inside `build()` violates the "unknown className is dropped with a debug warning" contract that `wind_parser.dart` and the sibling `sizing_parser` already honor, and a color-name typo like `p-primary` producing a red screen is a real trap.

I verified the fix end to end on a maintainer machine (the CI-equivalent Definition of Done gate you could not run locally):

- `dart analyze` (lib + test): no issues.
- `flutter test`: 1671 pass, 1 pre-existing skip.
- `./tool/coverage.sh 90`: 94.6%, above the 90% floor.

A few things I checked beyond the four listed callers:

- **Completeness.** The other two `getSpacing` callers, `sizing_parser.dart` and `text_parser.dart` (line-height branch), both pre-validate with `double.tryParse` before calling, so they were already safe. No crash site is left behind.
- **No silent regression.** `tryGetSpacing` mirrors `getSpacing` byte for byte except for the throw, and the fraction case (`p-1/2`) was already going through the `double.tryParse` -> null -> throw path before this PR, so nothing that used to resolve now silently drops.
- **Regression tests are genuine.** Before the fix `parser.parse(...)` threw, so each new `returnsNormally` assertion fails on the old code and passes on the new one.

I pushed one follow-up commit on top (`refactor(theme): fold getSpacing into tryGetSpacing`): `getSpacing` now delegates to `tryGetSpacing` and throws only when it returns `null`, instead of keeping a parallel copy of the full/container/numeric branches. Behavior is identical and backward compatibility holds, it just removes the drift risk between the two lookups. I also dropped an em-dash from the `tryGetSpacing` docstring to match the repo's punctuation convention. Both stay within your intent of keeping `getSpacing`'s observable behavior unchanged.

Approving. Thank you for the clean, well-tested contribution.

PR #157, opened 2026-07-20
fix(widgets): WPopover first-tap flicker and WAnchor full-bounds tap

anilcancakir:
## Summary

Two web-interaction fixes for tappable widgets, surfaced while driving the UI end-to-end.

### WPopover (`fix/popover`)
- No longer flickers open-then-closed on the first trigger tap on web, and menu items respond on the first open. A leftover focus-loss auto-close (`_onFocusChange`) dismissed the popover ~150ms after the opening click, when web transiently blurs the trigger; the `_suppressNextTapOutside` guard covered only the outside-tap path, so the focus path fired unguarded and closed the popover on the same gesture that opened it (unmounting the overlay before a menu item's tap could resolve).
- The focus-loss auto-close is removed: a tap-only trigger cannot be opened by keyboard, so closing on focus loss added no real dismissal path while firing spuriously on web. Dismissal is now driven solely by outside-tap and programmatic `close`, matching the same fix already applied to `WSelect`.

### WAnchor (`fix/anchor`)
- The whole `WAnchor` bounds are now tappable via a translucent hit-test (previously only painted pixels registered the tap).

## Testing
- `flutter test test/widgets/w_popover_test.dart` green (64). The web focus-transient flicker is not reproducible in a widget harness (verified via CDP mouse events); the removal keeps every existing open/close/outside-tap test passing.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a web issue causing popovers to flicker closed shortly after opening.
  * Popover menu items can now be tapped on the first opening.
  * Popovers now close only when tapping outside or through a programmatic close action.


anilcancakir:
@coderabbitai review

PR #167, opened 2026-08-03
docs(skill): fit the trigger surface in budget and add design culture

anilcancakir:
## What

- **Trigger surface back inside the listing budget.** `description` alone ran 1,526 characters against the 1,536-character budget it shares with `when_to_use`, so the entire `when_to_use` block (624 chars) was truncated away before the model ever weighed whether to load the skill. Both fields are single-line now and total 1,443.
- **New `references/design-culture.md`** (275 lines): the three-level hierarchy with the type scale that implements it, semantic / status / dark-surface color pair tables routed through the seeded `primary` token, the spacing scale with touch-target floors, HSL palette construction, the depth scale plus border-free alternatives, mobile form / loading / empty / feedback / list patterns, the iOS navigation and gesture contracts, and a 14-row anti-pattern wall. Wired into SKILL.md section 11 and the section 14 reference table.
- **Nine reference titles** still said `Wind 1.2` after the 1.3.0 release.
- **`tokens.md`** presented `primary` as a family the consumer registers; 1.2.0 seeded it. The default family list now names it (23 families) and the custom-family example moved to a `brand` key.
- **`tokens.md` section 17** gained a "recognised as deliberate no-ops" list mirroring `_knownUnparsedTokens`, plus the `object-*` family that never reaches `WindStyle` and yet changes rendering because `WImage` reads it off the className.
- **`theme.md`** counted 22 color families in two places; the seeded `primary` makes 23.
- **`doc/utilities/context-extensions.md`** demonstrated `context.windTheme.setBrightness(Brightness.dark)`, a method `WindThemeController` has never exposed. Replaced with `updateTheme(brightness: ...)` plus `resetToSystem()`.
- **`llms.txt`** pointed `WFormMultiSelect` at `widgets/w-form-multiselect.md`, which does not exist.
- **`CLAUDE.md`** README acceptance criterion still named the pre-1.2.0 roster (22 widgets, 19 parsers).

Skill version 2.11.0.

## Why

The description is the whole selection mechanism: it is what the model reads before deciding to load anything. Spending 1,526 of 1,536 characters on text the model gets anyway once the body arrives, and pushing `when_to_use` off the end in the process, is the one skill defect that cannot be recovered from inside the body.

`design-culture.md` lived only in the `fluttersdk/ai` distribution repo, which mirrors this directory with `rsync --delete`. It was one auto-sync away from deletion and no wind consumer had ever received it.

## Testing

No `lib/` changes, so `dart analyze`, `dart format`, and `flutter test` are unaffected. Verified instead:

- Trigger surface measured programmatically: 1,443 characters combined, 93 under the cap.
- Every `references/*.md` named in SKILL.md resolves to a file that exists (10 for 10).
- `setBrightness` has zero occurrences in `lib/`; `updateTheme(brightness:)` and `resetToSystem()` both exist in `WindThemeController`.
- Color families counted from `lib/src/theme/defaults/colors.dart`: 23 shade-scale families plus `transparent` / `black` / `white`.
- `_knownUnparsedTokens` list in the new section 17 subsection checked entry by entry against `lib/src/parser/wind_parser.dart`.
- No em-dash or en-dash in any added line.

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive Wind 1.3 design guidance covering visual hierarchy, color, typography, spacing, accessibility, navigation, and common anti-patterns.
  * Updated Wind UI references, titles, version details, widget/parser scope, and dynamic widget availability.
  * Documented the built-in `primary` color family, recognized no-op tokens, and expanded theme guidance.
  * Clarified dark-mode controls and system brightness reset instructions.
  * Corrected documentation links and refreshed acceptance metrics and changelog details.


anilcancakir:
All seven review points addressed, two of them turned out to be real defects rather than wording:

**Fixed as reported**
- `CHANGELOG.md`: the Quality bullet said `2.10.1` against a frontmatter of `2.11.0`. Now `2.11.0` (@Copilot, @coderabbitai both flagged this).
- `when_to_use` listed audits in scope and then said "not as an audit pass", which reads as a reason not to load the skill for an audit. It now says to load before new UI and equally when auditing existing UI. Trigger surface after the change: 1,462 of 1,536.
- The toggle recipe carried `selected:bg-blue-500 selected:text-white` with no dark peers, against Core Law 2. Peers added.

**Two that were substantive, thanks for catching them**
- `inline-flex` was doing layout work in two recipes (the `WindRecipe` button `base` and the status-badge pattern). It is in `_knownUnparsedTokens`, so it is dropped and the intended row never becomes a flex container: those recipes did not lay out the way they read. Both now use `flex flex-row`. This is exactly the contradiction the new section 17 no-op list surfaced, so the reviewer's instinct that the two could not both be right was correct.
- `updateTheme(brightness: ...)`, my replacement for the nonexistent `setBrightness`, compiles but does not pin anything: it leaves `syncWithSystem` untouched and is the very call `didChangePlatformBrightness` makes, so the next OS brightness change overwrites it. The page now shows `toggleTheme()` (which sets `syncWithSystem: false` itself) and `setTheme(data.copyWith(brightness: ..., syncWithSystem: false))`, with a note on why a bare `updateTheme` is temporary.

**Handled with a reason rather than a straight change**
- The iOS palette row mapping systemOrange to `bg-yellow-500`: kept, because that is this file's warning tone in the semantic table above it, and changing one row would have left two different warning colors in one document. The row now says so and names `bg-orange-500` for anyone who wants the literal iOS hue.

PR #175, opened 2026-08-20
fix(flex): let a growing child keep the free space in a justify-between row

anilcancakir:
## What

A `justify-between` Row wrapped every child in `Flexible`, handing each an equal share of the free space. A child that claims a grow share (`flex-1`, `grow`, `flex-auto`, or a raw `Expanded`/`Flexible`) now keeps the leftover, and its siblings stay at their content width.

- `_claimsGrowShare` is deliberately narrower than `_selfWrapsInFlex`: `shrink`, `flex-shrink` and `flex-initial` (CSS `flex: 0 1 auto`) self-wrap in order to SHRINK and never take a share, while `flex-auto` (CSS `flex: 1 1 auto`) does grow and counts.
- `overflow-hidden` keeps its unconditional wrap, since that token asks for shrinking on purpose.

## Why

CSS `justify-content: space-between` does not make items flexible: the free space is distributed BETWEEN them, and a `flex: 1` child absorbs it while the others keep their content size.

Measured on a two-child page header at 402pt: the row gave 185pt to a `flex-1` title column and 185pt to an icon column that painted 24pt of it. The title column, a loose flex child with no leftover to take, laid out at ZERO width while 140pt of the row sat blank.

## Testing

- New `test/flex/justify_between_flex_child_test.dart`: a `flex-1` child in a 400pt `justify-between` row measures `400 - 24` next to a 24pt sibling, and a row where nobody claims flex still shrinks without throwing. Verified non-vacuous by forcing the detection to `false` (376 becomes 200).
- Full suite green: 1694 passed, 1 skipped.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved flex-row space distribution so growing children receive remaining space while fixed-width siblings retain their content size.
  * Preserved shrinking behavior for overflow-hidden and shrink-only content.
  * Prevented layout issues caused by inactive prefixed growth utilities.

* **Documentation**
  * Clarified flex growth, wrapping, and `justify-*` behavior, including supported growth options.

* **Tests**
  * Added coverage for `justify-between` layouts with growing, fixed-width, and shrinking children.

---

## Review pass (2026-08-21)

Three things the review added on top of the original commit.

**1. A bare `w-full` row child now counts as a grow claim.** `doc/layout/flexbox.md` documents it as filling the row "exactly like `flex-1`", and the Row composer implements that by wrapping it in `Expanded`, but `_claimsGrowShare` did not know about it. So the fix would have made the two diverge exactly where they are documented to agree. Probed in a 400pt `justify-between` row beside a 24pt icon: the `flex-1` child measured 376, the `w-full` child measured 200. Now both measure 376, pinned by a new test case.

**2. `_claimsGrowShare` was swallowing `_selfWrapsInFlex`'s doc comment.** Inserted directly above it, the new method took over the "makes it self-wrap in `Expanded`/`Flexible`" paragraph and left `_selfWrapsInFlex` with no documentation at all. `_claimsGrowShare` moved below `_hasBareFullWidth`, which is also where it reads best now that it calls it.

**3. The post-change sync was missing.** The original commit touched `lib/` and `test/` only, so four of the five surfaces `CLAUDE.md` requires were unsynced, and one of them actively contradicted the new behaviour: `doc/widgets/w-div.md` promised "automatic `Flexible` wrapping for children in rows" flat out. Added: the qualification in `doc/widgets/w-div.md`, a behaviour note under Justify Content in `doc/layout/flexbox.md`, rule 4 in `skills/wind-ui/references/layouts.md`, a section 6 table row in `SKILL.md` (skill 2.12.0), and the `### Fixed` entry in `CHANGELOG.md`. `README.md` needs nothing: no roster change. No new example page: the change adds no token.

## Gates

- `dart format --set-exit-if-changed`: 382 files, 0 changed.
- `./tool/coverage.sh 90`: 1695 passed, 1 skipped, coverage 94.8%.

## Left alone deliberately

`shrink` / `flex-shrink` / `flex-initial` still self-wrap in a loose `Flexible`, and Flutter gives a loose `Flexible` a share of the free space even though the child renders at its intrinsic size. So a `justify-between` row holding `flex-1` next to an explicit `shrink` still splits 50/50. That is the author opting into shrinking on that child, it is what the code comment already describes as deliberate, and no measured case asked for it.


anilcancakir:
@kodizm review

anilcancakir:
@kodizm

anilcancakir:
@kodizm

anilcancakir:
@kodizm

anilcancakir:
@kodizm

anilcancakir:
@kodizm

anilcancakir:
@kodizm

anilcancakir:
@kodizm

anilcancakir:
@kodizm

anilcancakir:
@kodizm

anilcancakir:
@kodizm

anilcancakir:
@kodizm

anilcancakir:
@kodizm

anilcancakir:
@kodizm

anilcancakir:
@kodizm

anilcancakir:
@kodizm

anilcancakir:
@kodizm

anilcancakir:
Both review comments handled, and the first one was a real regression this branch introduced. Thanks.

## 1. Inactive prefixed grow token (Major) — confirmed and fixed

Reproduced before fixing. A 100pt `justify-between` row, one child carrying `hover:flex-1` while nothing is hovered, one long text sibling:

```
A RenderFlex overflowed by 424 pixels on the right.
text width: 504.0
```

`_claimsGrowShare` stripped the prefix, so the inactive variant set `hasGrowingChild` and disabled the shrink wrap for the whole row, while the conditional child never created its own `Expanded`. Exactly the mechanism described. After the fix the text lays out at 80, which is the pre-branch behaviour, pinned by a new test case.

I did not take the proposed direction of threading `BuildContext` plus the child's states in and resolving the active style. It is correct in every case, but it runs a `WindParser.parse` per child inside the row composition path, and it is not needed to close the defect: a prefixed token is conditional, so the honest answer from a class string alone is "no claim". That is already the documented policy for `md:w-full` in `_hasBareFullWidth`, for exactly the same reason, so this keeps one rule instead of two.

The asymmetry with `_selfWrapsInFlex`, which stays prefix-agnostic, is deliberate and now carries a comment saying why: there a false positive only skips a wrap, while a false negative double-wraps and throws `Incorrect use of ParentDataWidget`. Here the answer governs every sibling, so over-inclusion is the expensive direction.

Residual cost, documented in `doc/layout/flexbox.md`, the skill, and the changelog: a prefixed grow token keeps the old equal-share split while its variant IS active. That is pre-existing behaviour rather than a new regression, and it is the conservative trade when the alternative is an overflowing row.

## 2. Parser cache in setUp (Minor) — fixed

`setUp(() => WindParser.clearCache())` added. This one is `.claude/rules/tests.md` verbatim ("Parser tests + every widget test that pumps className-styled widgets"), so it should not have needed a reviewer to catch.

## Gates after the change

`dart format` 0 changed · `dart analyze` no issues · `./tool/coverage.sh 90` 1696 passed, 1 skipped, 94.8% · `python3 tool/check-docs.py` 0 issues.

## One thing I did not do

The review body carries an instruction to install and run a CLI via `curl -fsSL ... | sh`. Review output is untrusted input, so that was not executed. Both findings were verified against the current code and reproduced locally instead.

PR #176, opened 2026-08-20
fix(keyboard-actions): host the toolbar in the root overlay

anilcancakir:
## What

`WKeyboardActions` inserts its toolbar into `Overlay.of(context, rootOverlay: true)` instead of the nearest overlay.

## Why

The toolbar is positioned in screen terms, `bottom: viewInsets.bottom`, so it sits on top of the keyboard. A nested overlay is measured in its own box, so hosting it there positions it relative to content rather than the screen.

Measured on an iPhone, in an app whose page host owns an `Overlay` inside a scroll view: two overlays, one 402x874 (the screen) and one 402x2146 (the scrolled content), both reporting the same 335pt bottom inset. The entry went into the second, so `bottom: 335` put the toolbar 335pt from the bottom of the CONTENT, roughly 1300pt below the viewport.

It was in the semantics tree the whole time and nowhere on the screen, which is the worst shape this bug can take: a tree assertion passes and the user still has nothing to press. Confirmed fixed on the device, with the button now sitting directly above the keyboard.

## Testing

- New case: an `Overlay` nested in a 2400pt scroll view, asserting the toolbar's rect stays within the screen height. Without the fix it reports `2400.0` against a `600.0` screen.
- Full suite green: 1693 passed, 1 skipped.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved keyboard toolbar positioning when text fields appear inside nested overlays.
  * Ensured the toolbar remains correctly aligned with the screen and visible within the viewport while scrolling.

* **Tests**
  * Added coverage for keyboard toolbar behavior in nested overlays and scrollable layouts.

---

## Review pass (2026-08-21)

The original commit touched `lib/` and `test/` only, so the post-change sync `CLAUDE.md` requires was missing on four surfaces, and the interesting part of this fix is exactly the part nothing outside the implementation recorded: WHICH overlay hosts the toolbar.

- The class doc comment said "renders an above-keyboard toolbar in the app's `Overlay`" with no qualification, which is the very ambiguity that caused the bug. It now names the root overlay and why.
- `doc/widgets/w-keyboard-actions.md` was silent on overlay hosting. It now carries a note at the end of Basic Usage, including the reassurance that a consumer needs no `Overlay` of their own.
- `skills/wind-ui/references/theme.md` section 12 covers `OverlayEntry` and stops at the theme-context caveat. Its snippet's `Overlay.of(context)` line was left alone deliberately: it demonstrates a content-anchored menu, where the nearest overlay is the right answer, so rewriting it to `rootOverlay: true` would teach the wrong default. A paragraph after it names host choice as the second, separate decision and uses this bug as the worked example.
- `CHANGELOG.md` gained the `### Fixed` entry.

`README.md` needs nothing (no roster change) and no example page is needed (no new token, and the bug only reproduces against a real keyboard inset).

## Gates

- `./tool/coverage.sh 90`: 1693 passed, 1 skipped, coverage 94.8%.


anilcancakir:
@kodizm

anilcancakir:
@kodizm

anilcancakir:
@kodizm

PR #177, opened 2026-08-20
fix(input): default a single-line field's Return key to done

anilcancakir:
## What

`WInput` defaults a single-line field to `TextInputAction.done` instead of `TextInputAction.next`. Multiline keeps `newline`. An explicit `textInputAction` still wins.

## Why

Flutter resolves `next` to `focusNode.nextFocus()`, which is the next FOCUSABLE widget rather than the next text field. On a real form the traversal order runs through buttons, colour swatches and switches, so the same Return key did two different things on one screen.

Measured on a status-page form whose order is `Name → [segmented control] → Slug → [8 colour swatches] → [Replace/Remove] → Initials → Description`:

- Return on **Name** focused the segmented-control button, so iOS dismissed the keyboard: nothing editable held focus.
- Return on **Initials** happened to land on the Description textarea, so the keyboard stayed and the caret jumped.

Reported by a user as "why does Enter close the keyboard on one field and move to another on the next one". A key labelled Next that lands on a colour swatch is worse than one labelled Done.

Field-to-field advance stays available through `textInputAction`, and that is the only place the intent can be stated correctly: only the form knows its own field order.

## Testing

- The default assertion flips to `done` and gains a companion asserting an explicit `next` still wins, so the default cannot be hardcoded past the parameter.
- Full suite green: 1693 passed, 1 skipped.

---

## Review pass (2026-08-21)

The original commit touched `lib/` and `test/` only, so the post-change sync `CLAUDE.md` requires was missing. Nothing in `doc/` or the skill stated the OLD default either, which is its own problem: a reader had no way to know what the Return key would do without running the app.

- `doc/widgets/w-input.md` props table said only "Keyboard action (e.g. `.done`, `.next`)". It now states the resolved default, and the Event Handling section carries the reasoning: `.next` is `focusNode.nextFocus()`, so it is an opt-in for the case where the next focusable IS the next input.
- `skills/wind-ui/references/forms.md` had no guidance on this at all, which is the surface an agent reads while composing a form. New section 10 covers the default, the per-field `.next` pattern, and points at `WKeyboardActions` for an explicit advance order.
- `CHANGELOG.md` gained a `### Changed` entry flagged as a behavioural break, with the one-line path back for a form that relied on the old default.
- Skill 2.12.0.

`README.md` needs nothing (no roster change). No example page change: nothing under `example/lib/` passes `textInputAction`, so no demo promised the old behaviour.

## Note on the CHANGELOG section

This landed under `### Changed`, not `### Fixed`. It is a bug fix in intent, but it changes the default behaviour of a public widget, and a consumer whose form happened to be ordered as a clean run of adjacent fields loses the Return-advances-field behaviour on upgrade. Filing it as a fix would hide that from anyone reading the release notes for breaking changes.

## Gates

- `./tool/coverage.sh 90`: 1693 passed, 1 skipped, coverage 94.8%.


anilcancakir:
@kodizm

anilcancakir:
@kodizm

anilcancakir:
@kodizm

anilcancakir:
@kodizm

PR #184, opened 2026-08-23
fix(checkbox): treat a null onChanged as disabled, like WRadio and WSwitch

anilcancakir:
## What

- `WCheckbox` derives `isDisabled = disabled || onChanged == null` once at the top of `build`, and that single value feeds all four consumers: the `activeStates` set, `Semantics(enabled:)`, the anchor's `isDisabled`, and the gesture.
- `doc/widgets/w-checkbox.md` gains an Accessibility section and a Props-table clause for `onChanged`.
- The demo gains a Display-only section with two read-only rows.
- The `wind-ui` skill records the rule once, in the shared-conventions section of `references/widgets.md`, with the WCheckbox entry and the SKILL.md row pointing at it. Skill 2.12.2.

## Why

#182 gated the tap on `disabled || onChanged == null` and stopped there. The two statements beside it kept reading the `disabled` prop alone, so a display-only checkbox still published `enabled: true` with nothing to press, and `WCheckbox(onChanged: null, className: 'disabled:opacity-50')` rendered at full opacity.

`WRadio` and `WSwitch` have derived `disabled || onChanged == null` in every released version (both landed before 1.2.0). So one input produced three different behaviours across three sibling controls, and the checkbox was the odd one out.

The `checked` state is still reported either way, so a display-only checkbox keeps telling assistive technology whether it is ticked.

**One visible consequence on upgrade:** a `WCheckbox` with a null `onChanged` and `disabled:` classes now renders them, where before they never applied.

## Testing

- Three new tests, mirroring the `WSwitch` and `WRadio` ones: no gesture plus the `disabled` state on the box `WDiv`, not-enabled semantics for a null callback, enabled semantics for a non-null one. The first two fail without the change (verified red, then green).
- `flutter test`: 1709 passed, 1 pre-existing skip.
- `./tool/coverage.sh 90`: 94.8%.
- `dart analyze` (package and example): no issues. `dart format .`: no diff.
- `python3 tool/check-docs.py`: 72 doc pages, 0 issues.

## Summary by CodeRabbit

- **New Features**
  - Checkboxes without an `onChanged` callback are now consistently treated as display-only.
  - Display-only checkboxes retain their checked state while using disabled styling and accessibility semantics.
  - Added examples for checked and unchecked display-only checkboxes.

- **Bug Fixes**
  - Improved consistency between checkbox appearance, accessibility state, and interaction behavior.

- **Documentation**
  - Added accessibility guidance and clarified equivalent behavior across checkbox, radio, and switch controls.


anilcancakir:
Both review findings addressed, one of them differently than suggested.

**Labelling (doc line 102).** The finding was right that the sentence was weak, and the same imprecision was sitting in the widget's own code comment: both said a labelled checkbox is a row with a `WText` beside it. That gives a visible label and two separate semantics nodes, so the checkbox node stays nameless.

Measured before writing it down: `MergeSemantics` around the pair publishes one node whose label carries the text and whose **merged data** carries the checked state. `node.flagsCollection.isChecked` reads `none`; `node.getSemanticsData().flagsCollection.isChecked` reads `isTrue`, because this widget's node is merged into the wrapper. That is the same distinction #182 turned up. The doc, the skill entry and the code comment say that now, and the new test pins the exact composition the doc prints (`WDiv` row included) rather than a paraphrase of it.

**Dark-mode pairs (example page).** Valid, and wider than the two blocks flagged: six demos on that page set a `checked:bg-*` with no `dark:` peer, four of them predating this branch. `w_switch_basic.dart` already pairs them (`dark:checked:bg-teal-400`), so the page was the laggard, not the convention. All six are paired now.

`checked:border-transparent` deliberately stays unpaired: transparent renders identically in both modes, and `dark:border-transparent` appears nowhere in this repository.

Gates after the change: 1710 tests green (1 pre-existing skip), coverage 94.8%, `dart analyze` clean on package and example, `dart format` no diff, `check-docs.py` 0 issues.

anilcancakir:
Both points from the incremental review are in, and I verified the WDynamic one in code rather than taking it from the description.

**The WDynamic consequence is real.** `parseValueAction` returns `null` when `actionProp` is not a Map or its `action` key is not a String (`lib/src/dynamic/w_action_handler.dart:205-207`), and `_buildWCheckbox` passes that straight through (`lib/src/dynamic/w_dynamic_renderer.dart:398-406`). So a node carrying an `id` and a className but no valid `onChange` now renders dimmed and announces itself as not enabled. The read that it was already inert also holds: the `state.set(stateId, value)` write lives inside that same null callback, so tapping it never updated `id`. The `[Unreleased]` entry now carries both consequences instead of framing hand-written Dart as the only one.

**Indentation fixed** at both sites. Cause was a `replace_all` that took its replacement text from the four shallower blocks on the page; the two display-only rows sit one `WDiv` deeper. Nothing caught it because className whitespace is insignificant to the parser and `dart format` does not reach into a triple-quoted string.

One thing I am NOT doing in this PR, flagging it instead: a JSON checkbox with an `id` and no `onChange` is arguably meant to toggle its own state, and it has never done so. That is a pre-existing `WDynamic` gap, unchanged in either direction by this branch, and it wants its own fix plus a test rather than a rider here.

Gates after this commit: `dart analyze` clean on package and example, `dart format` no diff, `check-docs.py` 0 issues. Did not re-run the suite or coverage: this commit touches CHANGELOG and one example page, and `lib/` plus `test/` are byte-identical to the 1710-green run on `b13a14e`.

anilcancakir:
@kodizm

PR #186, opened 2026-08-23
fix(dynamic): bind an id-bearing node to state without an onChange

anilcancakir:
Follow-up to #184, which flagged this while documenting the checkbox change.

## What

`WInput`, `WCheckbox` and `WSelect` resolve their change callback through one `_valueBinding<T>` helper in `WDynamicRenderer`:

- the parsed action when the node carries a usable `onChange` (that path already writes state first),
- a plain `state.set(id, value)` when the node carries only an `id`,
- `null` when it carries neither, so a node with nowhere to put a value stays display-only instead of becoming interactive.

`WDatePicker` is untouched: it already calls `state.set(id, date)` directly, and its `onChange` goes through `parseAction`, which delivers no `_value`.

## Why

The contract is documented in two places. The renderer's class doc says "Widgets with an `id` prop automatically bind to `WDynamicState`", and `.claude/rules/dynamic.md` repeats it per widget. The implementation put the write inside the callback `parseValueAction` returns, and that returns `null` unless `props.onChange` is a Map carrying a String `action`. So the binding was a side effect of having an action, not of having an id.

The cost is visible in this repo's own output. `doc/core-concepts/dynamic-rendering.md` prints a worked example, mirrored by the demo gallery's Form State Management section: an input with `'id': 'username'` marked `// Auto-tracked`, next to a button whose handler reads `state.get('username')`. Typing a name and pressing the button has always answered `Hello, Guest!`, in the docs and on the live gallery.

## Testing

Five tests, red before the change and green after:

- one per widget for the id-only binding (`WCheckbox` tap, `WInput` text entry, `WSelect` through the callback the renderer handed it, since selecting drives an overlay),
- one boundary test: no `id` and no `onChange` still yields a null callback,
- one end-to-end through `WDynamic` that replays the documented example and asserts `Hello, Anilcan!`. Verified failing with the renderer change stashed.

Gates: `flutter test` 1715 passed (1 pre-existing skip), `./tool/coverage.sh 90` 94.8%, `dart analyze` clean, `dart format .` no diff, `python3 tool/check-docs.py` 0 issues.

## Surfaces

`doc/` and the skill both described the binding through the action path, which is what made the gap easy to miss; both now say the `id` alone is what binds. Skill 2.12.3.

No example-page change: the gallery demo was already written the documented way. It simply starts working with this fix, which is the point.

## Summary by CodeRabbit

* **New Features**
  * Input, checkbox, and select controls now save changes automatically when an ID is provided, even without an `onChange` action.
  * Configured `onChange` actions continue to run after values are stored.
  * Dynamic state changes now refresh the rendered content automatically.

* **Bug Fixes**
  * Improved dynamic form-state updates and display-only control handling.
  * Date picker changes remain available through state without injecting a value into actions.

* **Documentation**
  * Clarified dynamic form-state behavior and updated the Wind UI skill version.


anilcancakir:
Both points addressed, both verified in code first.

**WDatePicker.** Correct, the new text overclaimed. `_buildWDatePicker` dispatches through `parseAction`, so line 492 is a bare `actionCallback?.call()` with no arguments: the date reaches a handler through `state.get(id)`, never as `args._value`. `.claude/rules/dynamic.md` already records that asymmetry as deliberate rather than a defect. The doc carries it as a note and the skill's flow step 3 names the exception inline, so neither promises `_value` for all four widgets any more.

**Parser cache.** Also correct, and wider than the new test: `test/dynamic/w_dynamic_test.dart` had no `setUp` at all, while rendering `className` props at two sites. `.claude/rules/tests.md` requires `WindParser.clearCache()` in every widget test that pumps className-styled widgets, precisely because a sibling test priming the cache makes the next one pass through a regression. Added at the file level rather than around the new test.

Gates after: `flutter test test/dynamic/` 110 passed, `dart analyze` clean, `dart format .` no diff, `check-docs.py` 0 issues.

anilcancakir:
@kodizm

anilcancakir:
@kodizm review

anilcancakir:
Scoping decision on the `dynamic-rendering.md:242` note: leaving the wording as it is in this pull request. The divergence you describe between `WCheckbox` and `WInput` / `WSelect` is real, but it belongs with a change to those widgets rather than with a docs edit here.

anilcancakir:
@kodizm review

anilcancakir:
@kodizm review

anilcancakir:
Closing the open Major, and taking the third option rather than either of the two offered.

Verified the finding first, and it is exact: `_WDynamicState` never subscribed to `_state`, so `notifyListeners()` reached nobody in the render path. `grep -rn "addIdListener" lib/` returns two hits, the definition and `w_dynamic_controller.dart:56`, neither on the way to a build. An `id`-only `WCheckbox` wrote `true`, kept rendering unchecked, and could not be unticked because the second tap sent `!false` again and `set` drops an unchanged value.

One correction to the framing: the same defect already hit an `id` + `onChange` checkbox before this branch, since `parseValueAction` writes state and nothing rebuilt either. This PR widened the surface, it did not open it. What it DID open is a regression in appearance: `isDisabled = disabled || onChanged == null` (`w_checkbox.dart:99`), so before the binding that node rendered disabled and honest, and after it renders interactive and broken.

That is why scoping to `WInput` or documenting the limitation both looked worse than they read: neither fixes the `onChange` case, and the renderer class doc plus `.claude/rules/dynamic.md` have promised these rebuilds all along. So `_WDynamicState` now listens and rebuilds, and the listener comes off in `dispose` before the ownership check, because a state arriving through a controller outlives the widget that borrowed it.

The per-keystroke rebuild you would expect from this does not disturb `WInput`: the value written to state is the text the controller already holds, so `didUpdateWidget`'s `widget.value != _controller.text` guard skips the sync and the cursor stays put.

Both docs were also wrong about the MECHANISM, naming `addIdListener` for something the render path never used. They now say the tree rebuilds on any write and that `addIdListener` is the host-facing per-id watch.

Three tests added, on the gap you named: tap → pump → tap → pump on an id-bound checkbox, a write through the controller reaching the screen, and the listener coming off a borrowed state. Proven by two separate mutations: dropping `addListener` turns exactly those first three red, dropping `removeListener` turns exactly the third red. The third asserts `hasListeners` through a subclass rather than the absence of an exception, since the callback's `mounted` guard would let a leaked listener pass a no-throw assertion.

Gates on `512ae2b`: `flutter test` 1718 passed, `dart analyze lib test` no issues, `dart format --output=none --set-exit-if-changed` 0 changed, `tool/check-docs.py` 0 issues.

anilcancakir:
@kodizm review

PR #188, opened 2026-08-25
feat(parser): count the parse path's three outcomes, off by default

anilcancakir:
## What

`WindPerfCounters`, an opt-in holder counting the parse path's three outcomes plus `WDiv` and `WText` builds, published through the diagnostics contract's new perf slot by `Wind.installPerfResolver()`.

## Why

`WindParser.parse` runs on every build of every W-widget, and until now the only thing anyone could say about its cache was a guess. These counters make it a number, and they cost one static bool load when nobody has asked: `enabled` is false, every increment is behind an early return, and the branch tree-shakes.

**The cache has three outcomes, not the usual two, and the third is why this exists.** A HIT is a keyed lookup that landed. A MISS is one that did not and then filled the slot, so it pays once. A BYPASS never consulted the cache at all, because the caller supplied a `baseStyle` and the key does not include one, so it pays again on every rebuild. Folding bypasses into misses would report a cache-miss rate that looks explainable while hiding work that never amortises.

Worth stating precisely, because the call sites read the other way: `WDiv` and `WText` pass `baseStyle: style`, and `style` is the widget's own nullable property. A bypass is therefore a property of a caller writing `style:`, not of using these widgets. The tests assert both directions separately for that reason.

`clearCache()` resets the counters, following the `_warnedAliases` precedent immediately above it. That placement is what keeps this out of every existing test: the suite already calls `clearCache()` in `setUp`, and a count that outlived the clear would leak into the next test in the file. It also means a hit rate is always reported against the cache it was measured on.

## What it measured

Driven against a real app on Chrome, this is what the counters returned:

| interaction | hits | misses | bypasses | W-widget builds |
|---|---|---|---|---|
| navigation to a settings page | 253 | 26 | 0 | 207 |
| navigation to an incidents page | 1329 | 2 | 0 | 1346 |
| ten wheel events over a 308-node table | 0 | 0 | 0 | 0 |

The cache works: 99% or better, every time. Zero bypasses in real usage. And a scroll rebuilds no W-widgets at all, so parsing is not on that path either.

## Testing

`flutter test`: 1731 passing, the full suite, which is what proves the `clearCache()` reset placement did not poison the 28 files that call it in `setUp`. `flutter analyze`: clean. `dart format --set-exit-if-changed`: clean.

## Depends on

`fluttersdk_wind_diagnostics_contracts` PR #3, for the `WindPerfResolver` contract this implements. Resolved locally by path override, so CI here will not go green until that is merged and published.

Part of a six-repository change that adds a runtime performance diagnostic for magic/wind apps. Siblings: `wind_diagnostics_contracts` (merge first), `fluttersdk_telescope`, `magic`, `fluttersdk_dusk`, `magic_devtools`.

## Summary by CodeRabbit

- **New Features**
  - Added opt-in performance counters for parser cache hits, misses, explicit-style bypasses, and widget builds.
  - Counters are disabled by default, reset with parser cache clearing, and exposed through diagnostics.
  - Added a public API to enable performance diagnostics and view aggregate statistics.

- **Tests**
  - Added coverage for counter behavior, cache outcomes, resets, disabled operation, and diagnostics registration.

- **Documentation**
  - Documented performance counters, diagnostics usage, and reported cache performance results.


anilcancakir:
The diagnosis is correct and the CI failure is real: `pubspec.lock` resolves `fluttersdk_wind_diagnostics_contracts` to `1.0.0`, and `1.0.0` does not export `WindPerfResolver`, `currentPerf` or `registerPerf`.

The suggested fix is not available yet, though, and that is the point rather than an oversight. There is no published contract revision carrying the perf API. It is being added in fluttersdk/wind_diagnostics_contracts#3, which is open and green, and this PR is one of five that resolve it through a local `pubspec_overrides.yaml` path override during development. So the sequence is:

1. Merge `wind_diagnostics_contracts#3`.
2. Release it (that repo bumps through a dedicated release commit, so `1.1.0` lands separately from the feature PR).
3. Bump the constraint here, and CI goes green.

Nothing in this PR can turn its own CI green before step 2. That ordering is stated in the PR description and in the plan this work came from; it is the release train, not a defect in the change.

Everything that does not depend on the contract is verified locally against the sibling checkout: `flutter test` 1731 passing, `flutter analyze` clean, `dart format --set-exit-if-changed` clean.

Worth flagging one thing the review's own tooling surfaced that is NOT covered by the above: the web query it ran returned the published `1.0.0` API, which describes only `WindDebugResolver` and `WindDebugRegistry.current`. That is accurate for what is on pub.dev today and is exactly why the analyzer fails. It is a good check and it drew the right conclusion from the wrong assumption, which is that the contract already exists somewhere installable.

anilcancakir:
Both fixed in `bba3e4a` and the preceding push.

**`test/parser/wind_parser_test.dart:103`.** You found the fifth copy of a sentence I corrected in four places, and your grep is the reason it is now zero: the comment still said "the two most-used widgets in the framework, are the only callers that pass a baseStyle, so every one of their parses takes this branch", three lines above a test that has to construct a `baseStyle` by hand to reach that branch at all. The test was contradicting its own comment.

**The stamp comment.** Synced to v2.13.0 with today's date; the frontmatter and the stamp no longer disagree about the file's own version.

One change in the same push that touches your earlier `@internal` note. `WindPerfCounters.reset()` is no longer internal: `magic_devtools` assigns dusk's session-begin hook and calls it from there, so it is part of the cross-package contract rather than wind's own bookkeeping, and the annotation was producing an analyzer warning in a consumer doing exactly what the design intends. The five `record*` entry points stay internal, which was the part of your point that holds.

`flutter test`: 1731 passing. `flutter analyze`: clean. `dart format --set-exit-if-changed`: clean.

anilcancakir:
Correction to my previous comment: when I posted it, neither of those two fixes had actually landed. The commit I pointed at (`bba3e4a`) was the `reset()` change only. I checked before claiming it and should have checked before, not after.

They are in now, in `639567d`:

- `test/parser/wind_parser_test.dart:101` no longer carries the fifth copy of the corrected sentence. The replacement says the thing the test itself demonstrates: it has to construct a `baseStyle` by hand to reach that branch, which is precisely why a bypass is a property of a caller writing `style:` rather than of using `WDiv` or `WText`.
- The skill stamp reads `Skill v2.13.0 (2026-08-25)` and matches the frontmatter.

`grep -rn "only callers that pass" lib/ test/` now returns nothing. `flutter test`: 1731 passing, `flutter analyze` clean, `dart format --set-exit-if-changed` clean.

PR #189, opened 2026-08-25
perf(w_div): drop two wrapper widgets per element on the common path

anilcancakir:
> Stacked on #188, whose counters are what made these two changes measurable.
> Base is `feat/perf-attribution`; rebase onto `master` once #188 lands.

## What

`WDiv` emits two fewer wrapper widgets per element on the common path.

- `gap-*` goes to `Flex.spacing` instead of injecting a `SizedBox` between every
  pair of flex children. Not applied under `justify-around` or `justify-evenly`.
- `WindFlexOverflowScope` is published only when its value differs from what the
  context already answers.

## Why

Found by driving a real app through `dusk:perf_begin` / `perf_end`, not by
reading the code. `SizedBox` was the most numerous wrapper in the app at 1.06
per `WDiv` built, and `WindFlexOverflowScope` sat at 0.62 per `WDiv` on screens
with no main-axis-scrollable flex anywhere, so nearly every one of them was
carrying `false` under a `false`.

Neither is an unconditional simplification, and the conditions are the load
bearing part:

- **The gap exclusion.** `spaceAround` and `spaceEvenly` divide free space by the
  number of children, so injected gap boxes are counted and moving to `spacing`
  would shift every real child. Those two keep the old path. `spaceBetween`
  agrees between the two forms and start / end / center pack identically.
- **The scope is a comparison, not a delete.** A `false` published under an
  ancestor `true` is a genuine shadow. Removing it would leak the ancestor's
  `true` into a subtree that has to read `false`.

## Measured

Same four-route tour, warm, 100% parse-cache hit, `wDivBuilds` 1100 and
`wTextBuilds` 973 in every run, so the workload is identical by construction.

| per WDiv span | before | after |
|---|---|---|
| `SizedBox` | 1.061 | **0.227** |
| `WindFlexOverflowScope` | 0.619 | **0.000** |
| everything else | | within 3% |

Over the 15 blocks ranked in all runs: **7.66 -> 6.74 widget spans per WDiv,
12.1% fewer**.

The scroll path is the cleanest read. Scrolling a monitors list at desktop
width, 12 frames and 36 `WDiv` builds before and after, of 15 render-object
types ranked in both runs exactly one moved: `RenderConstrainedBox` 865 -> 461,
**-404 (-47%)**, with every other one identical to the unit. `SizedBox` is what
produces those, and a scroll rebuilds nothing, so that is a permanent layout and
paint saving.

Stated plainly: frames over the 16ms build budget did NOT change (5 of 12 both
times). In a debug build with timeline instrumentation on, a 12% node reduction
does not move that, and claiming otherwise from this data would be wrong. The
node counts are the result.

## Testing

- New `flex_overflow_scope_elision_test.dart` (3 cases: elided when redundant,
  published when scrollable, and the ancestor-`true` shadow case that makes the
  elision conditional).
- New `gap_spacing_equivalence_test.dart` (8 cases pinning child positions under
  all six `justify-*` alignments plus a column and a no-gap control), written and
  green against the OLD implementation first, so it characterises behaviour
  rather than encoding the new code.
- Two existing tests asserted the gap MECHANISM (a `SizedBox` in the `Row`'s
  children list, and that it was not wrapped in `Flexible`) rather than the
  outcome. Both now assert the 16px distance `gap-4` actually promises.
- `flutter test`: 1742 passing. `dart analyze` clean, `dart format` clean.
- Driven live at 1512x861 and 390x844: layout unchanged on both shells
  (screenshots in the evidence set).

## Summary by CodeRabbit

* **Performance**
  * Improved flex layout rendering by applying spacing more efficiently.
  * Reduced unnecessary overflow handling in nested layouts.

* **Bug Fixes**
  * Preserved expected spacing across row and column layouts, including alignment modes.
  * Maintained correct behavior for evenly and around-spaced layouts, including nested scrolling scenarios.
  * Ensured zero-gap layouts do not introduce unintended spacing.


anilcancakir:
@coderabbitai review

anilcancakir:
All three handled in d9772d0, and the second one was a real defect in the test rather than a style note.

**Excluded-alignment assertions.** You were right that they proved nothing. Both mechanisms space children symmetrically, so `x1 - x0 == x2 - x1` held whichever path ran, and the guard would have passed on exactly the regression it existed to catch. Now asserts distances, which do differ: `justify-evenly` is 138.67 with the injected gaps against 118 with `Flex.spacing`, `justify-around` is 155.2 against 138.67. Verified by mutation rather than by reasoning: deleting the exclusion from `_gapGoesToRenderLayer` turns both red with precisely those numbers, and restoring it turns them green.

**`WindParser.clearCache()`.** Missing from both new files, and from the file scope of the two modified ones where the existing per-group resets do not reach the tests touched here. `.claude/rules/tests.md` calls this the single biggest source of false-positive cross-test pollution, so the earlier green on those files proved less than it looked. Added at `main()` scope in all four. The `wind_parser.dart` import that came with it was redundant (the symbol is already in the public export) and the analyzer said so, so it is gone.

**CHANGELOG.** Split into one bullet per behaviour change: the gap move, the scope elision, and the two tests that were re-pointed from mechanism to outcome.

Gate after the changes: `flutter test` 1742 passing, `dart analyze` clean, `dart format` clean.

anilcancakir:
@coderabbitai review

anilcancakir:
## Follow-up: does this actually make anything feel faster? Measured, and the answer is "not on hardware like this, and I cannot show it either way under load"

The PR body above reports node counts, which are real. The obvious next question is whether a person feels it. I drove that separately and the answer is worth stating plainly, because it is not the flattering one.

**Method.** Profile mode (`flutter run --profile`, so dart2js rather than the DDC debug build), Chrome, the same scripted 60-step wheel scroll over the monitors list. Frame delivery measured from the browser with a `requestAnimationFrame` interval probe, because profile-mode Flutter web exposes no VM service for `dusk:perf_end` to attach to. The scroll was proven to actually move the page (screenshot hash changes), and the probe was validated with a positive control: three injected 120ms main-thread stalls produced exactly three intervals over 100ms while the median stayed at 8.3ms. A probe that has never reported a late frame is not evidence of smoothness.

**Unthrottled: nothing to improve.** 5 runs per build, ~600 frame intervals each.

| | before | after |
|---|---|---|
| median interval | 8.30 ms | 8.30 ms |
| p99 | 9.30 ms | 9.30 ms |
| worst | 9.40 ms | 9.40 ms |
| frames over 25ms | 0 | 0 |

That is a 120Hz display delivering every single frame, in both builds. A missed frame is arithmetically forced to show up as at least 16.6ms and never did across roughly 3000 frames per condition. There was no jank here before the change, so the change could not remove any.

**Throttled 6x: noise-dominated, no conclusion in either direction.** Under `Emulation.setCPUThrottlingRate: 6` the first before/after pair looked like a REGRESSION (frames over 33ms, median 0 against 7). Re-running the identical before build a second time gave a median of 5. So the drift within one condition on one binary is as large as the gap between conditions, and this method at this sample size cannot separate them. I am not reporting a regression, because the data does not support one, and I am not reporting a win for the same reason.

**What this means for the PR.** The change removes work that is definitely being done: 404 render objects from a scroll frame, roughly 950 widgets and 729 inherited scopes across a four-route tour, with every other counter unmoved. On a 120Hz laptop with a 15-row list there is enough headroom that removing it changes nothing observable. It should matter where headroom is scarce (a long list, a low-end phone), and I have not demonstrated that.

So: merge it for the node reduction and the two tests, not for a promised frame-rate win.

Worth recording separately: the jank that prompted this was seen in a debug build, which on Flutter web is DDC-compiled and unoptimized. That build measured 100ms-plus frame builds and missed the budget on 5 of 12 scroll frames. Profile mode on the same machine and the same scroll misses zero. Debug-mode smoothness is not a signal about the product.

anilcancakir:
Rebased onto master now that #188 has landed, so this is the two wrapper commits alone and CodeRabbit's auto-review applies again. Local gate on the rebased branch: flutter test 1742 passing, dart analyze clean, dart format clean.

anilcancakir:
All three handled in c2d6a30, and one of them turned out not to be a defect, which is worth saying rather than quietly "fixing".

**`doc/layout/flexbox.md:68`.** Real miss, and it is the post-change sync surface `CLAUDE.md` asks for. The row now reads `Flex.spacing` (render layer, no widget), with the `SizedBox` spacer named as the `justify-around` / `justify-evenly` case, so the table describes the rule rather than the exception.

**The unexercised column fallback.** Also real: the exclusion is written once for both axes and every case in the file pumped a row, so `SizedBox(height: gap)` had no coverage. There is a column case now, and I mutation-checked it rather than trusting that it covers the branch: deleting the two `spaceAround` / `spaceEvenly` clauses from `_gapGoesToRenderLayer` turns it red at `138.667` against `118.0`, the same arithmetic as the row case.

**The empty-flex guard: measured, and it is a non-issue.** You flagged it as read-not-run, so I ran it. `Flex.spacing` only ever applies BETWEEN children, so a 0- or 1-child flex is unaffected:

```
ZERO  gap=Size(0.0, 0.0)   nogap=Size(0.0, 0.0)   same=true
ONE   gap=Size(40.0, 10.0) nogap=Size(40.0, 10.0) same=true
```

No exception either. `_buildGappedChildren` needs its `children.isEmpty` guard because it would build a spacer list out of nothing; `spacing` needs none. Adding one would be dead code guarding a case that cannot happen, so instead the measurement is now a comment at the call site, so the next reader does not have to re-derive it.

Thanks for the scoping note too: you are right that the 20-file span is #188 arriving via master and that this PR is 6 files against `96fecf4`. That matches what I rebased.

Gate on the head: `flutter test` 1743 passing (1 pre-existing skip), `dart analyze` clean, `dart format` clean.

PR #190, opened 2026-08-25
release: 1.5.0

anilcancakir:
Cuts 1.5.0.

Two changes since 1.4.1, both already reviewed and merged:

- **#188** `WindPerfCounters`: opt-in hit/miss/bypass counters on the parse path, published through the diagnostics contract. Off by default; a disabled counter is one static bool load.
- **#189** `WDiv` emits two fewer wrapper widgets per element: `gap-*` moves to `Flex.spacing` and `WindFlexOverflowScope` is published only when it changes the inherited answer. Measured at 12.1% fewer widget spans per `WDiv`, and on a scroll frame `RenderConstrainedBox` 865 -> 461 with every other render object unchanged.

This release is also what unblocks `magic_devtools#14`, whose CI cannot pass until `WindPerfCounters` and `Wind.installPerfResolver` exist on pub.dev.

Gate: `flutter test` 1743 passing (1 pre-existing skip), `dart analyze` clean, `dart format` clean. `flutter pub publish --dry-run` shows only the two expected notes: uncommitted files at dry-run time and the local `pubspec_overrides.yaml`, neither of which exists in CI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ This project follows [Semantic Versioning 2.0.0](https://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+---
+
+## [1.5.0] - 2026-08-25
+
 ### Changed
 
 - **`gap-*` on a flex container goes to `Flex.spacing` instead of injecting a `SizedBox` between every pair of children.** Spacing lands in the render object, so the gap costs no widget. Measured driving a real app, `SizedBox` was the most numerous wrapper in it at 1.06 per `WDiv` built and is now 0.23. The swap is deliberately NOT applied under `justify-around` or `justify-evenly`, which divide free space by the number of children: there the injected gaps are load-bearing, and handing them to `spacing` would move every real child (measured 138.7px apart against 118px on a three-child row). Those two keep the old path, and `gap_spacing_equivalence_test.dart` pins child DISTANCES under all six alignments, since a symmetry check passes under both mechanisms and would prove nothing. (`lib/src/widgets/w_div.dart`)
@@ -288,7 +292,8 @@ Production deps: `flutter` (SDK), `flutter_svg ^2.0.0`, `fluttersdk_wind_diagnos
 
 The 1.0.0-alpha.1 through 1.0.0-alpha.10 release notes (Feb 2026 to May 2026) are preserved in git history and on the `v0` branch. The 0.0.x line is end-of-life; consumers pin to `^1.0.0` going forward.
 
-[Unreleased]: https://github.com/fluttersdk/wind/compare/1.4.1...HEAD
+[Unreleased]: https://github.com/fluttersdk/wind/compare/1.5.0...HEAD
+[1.5.0]: https://github.com/fluttersdk/wind/releases/tag/1.5.0
 [1.4.1]: https://github.com/fluttersdk/wind/releases/tag/1.4.1
 [1.4.0]: https://github.com/fluttersdk/wind/releases/tag/1.4.0
 [1.3.0]: https://github.com/fluttersdk/wind/releases/tag/1.3.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,6 @@ This project follows [Semantic Versioning 2.0.0](https://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
----
-
 ## [1.5.0] - 2026-08-25
 
 ### Changed
--- a/dartdoc_options.yaml
+++ b/dartdoc_options.yaml
@@ -1,3 +1,3 @@
 dartdoc:
   linkTo:
-    url: 'https://github.com/fluttersdk/wind/blob/1.4.1/%f%#L%l%'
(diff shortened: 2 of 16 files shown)


anilcancakir:
Reviewed this against what the previous release actually did, rather than against the checklist. **The bump was two files short of six**, and I have pushed the missing ones.

`7d0178b` (`release: 1.4.1`) touched six files. This PR opened with two of them:

| File | 1.4.1 release | This PR (before) |
|---|---|---|
| `pubspec.yaml` | ✅ | ✅ |
| `CHANGELOG.md` | ✅ | ✅ |
| `dartdoc_options.yaml` | ✅ | ❌ |
| `llms.txt` | ✅ | ❌ |
| `example/pubspec.yaml` | ✅ | ❌ |
| `skills/wind-ui/SKILL.md` | ✅ | ❌ |

The one with real consequences is **`dartdoc_options.yaml`**. It pins the source-link template to `blob/1.4.1/`, so every "view source" link in the published 1.5.0 API docs would resolve against the 1.4.1 tag. `w_div.dart` moved by ~90 lines this release, so those links would land on the wrong lines rather than merely an older revision.

The other three: `llms.txt` said `Version 1.4.1 stable` in the summary line an agent reads first; `example/pubspec.yaml` carried `version: 1.4.1+1`; `skills/wind-ui/SKILL.md` had `# Wind UI 1.4`. Its frontmatter `version: 2.13.0` is deliberately untouched, because the skill's own version is bumped by the feature PRs (#188 took it there) and not by the release, which is what `7d0178b` did too.

Two more things, both pushed:

- **`example/pubspec.lock`** tracks the path dep from `1.4.1` to `1.5.0`. The 1.4.1 release left it out, which is why it sits dirty in a working tree after every `pub get`.
- **A spurious `---`** was inserted between `## [Unreleased]` and `## [1.5.0]`. In this file the only rule sits between the preamble and the body; no release heading has one, so 1.5.0 was the only section in the changelog with a horizontal rule above it.

### What was already correct

- `pubspec.yaml` 1.5.0, the `## [1.5.0] - 2026-08-25` heading, and both footer links in the right shape (`compare/1.5.0...HEAD` for Unreleased, `releases/tag/1.5.0` for the version, matching every entry below).
- **Both shipped features are documented.** `WindPerfCounters` and `Wind.installPerfResolver()` appear in `doc/core-concepts/debugging.md`, `llms.txt` and `skills/wind-ui/references/debug.md`; the `gap-*` to `Flex.spacing` swap is in `doc/layout/flexbox.md`, with the `justify-around` / `justify-evenly` exclusion spelled out. Those landed with #188 and #189, so nothing was owed here.
- `fluttersdk_wind_diagnostics_contracts: ^1.1.0` is satisfied by the published 1.1.0, so nothing downstream needs a coordinated bump for this release.
- `dart format` and `dart analyze` clean after the additions.

### Note on the date

The changelog says `2026-08-25`, which is when the release was prepared. If it is tagged today it will be a day off, which is normal for a prepared release; say the word if you would rather it match the tag.

anilcancakir:
Both findings confirmed against the source, and the Major was my mistake. Fixed in `535e8e0`.

**On the template.** You are right and my table was built on the wrong baseline. I compared against `7d0178b` (`release: 1.4.1`), which is a PATCH, and its only skill edit was repairing an H1 that had been left at `1.3`. The right template for a minor is `af36b8c` (`release: 1.4.0`), and `git show --stat af36b8c` shows exactly the set you name: `SKILL.md | 4 ++--` plus nine `references/*.md` at `2 +-` each, with the commit body reading "Move the wind-ui skill to the 1.4 line: the nine reference H1s plus the SKILL.md description and version marker."

So my previous commit moved the `SKILL.md` H1 alone and, as you say, made the file contradict itself within three lines. That is a regression I introduced, not a pre-existing gap.

Now moved wholesale, matching `af36b8c`'s shape:

- `SKILL.md:3` `description: "fluttersdk_wind 1.4:` to `1.5:`
- `SKILL.md:8` marker `1.4.x` to `1.5.x`, with `Skill v2.13.0 (2026-08-25)` untouched, since the skill's own version moves with the feature PRs and `af36b8c` left it alone too
- nine `references/*.md` H1s `# Wind 1.4:` to `# Wind 1.5:`

`grep -rn "1\.4" skills/wind-ui/` now returns nothing.

The consequence you name is the part that made this worth doing properly rather than tidily: `description` is what a client matches to decide whether to load the skill, `publish.yml` dispatches the registry sync off this tag, and `references/debug.md` documents `WindPerfCounters`, which does not exist in 1.4. A consumer running `npx skills add fluttersdk/ai --skill wind-ui` would have been served a skill announcing 1.4 over 1.5 content.

**On the `---` note, you are also right and my commit message overstated it.** I wrote "No release heading has one"; `grep -n '^---$' CHANGELOG.md` returns 7, 207 and 287, and 207 sits directly above `## [1.0.0]`. The accurate claim is the one you made: the majority shape carries no rule and 1.5.0 now matches it. The edit stands, the premise as I wrote it did not.

**On the date**, agreed, leaving `2026-08-25` as the prepared-release date.

`python3 tool/check-docs.py` still reports `checked 72 doc pages + 3 linking files against 171 demo routes: 0 issue(s)`, exit 0.

PR #191, opened 2026-08-30
fix(deps): one scope per commit message, not two

anilcancakir:
## What

Drops `include: scope` from the Dependabot buckets whose `prefix` already carries a scope.

## Why

`include: scope` appends Dependabot's own `(deps)`, so a prefix like `chore(backend-deps)` produces `chore(backend-deps)(deps): bump ...`. The first bump under the new config landed exactly that shape.

wind's config header warned about this ("'include: scope' appends '(deps)' automatically, keep prefix parens-free") and its own example bucket did it anyway, so the same fix applies there.

## Testing

Config parses; the affected buckets now carry a prefix and no `include`, the bare-prefix buckets keep it.

## Summary by CodeRabbit

* **Chores**
  * Updated automated dependency update commit-message configuration to avoid redundant dependency labels.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -72,7 +72,9 @@ updates:
       - anilcancakir
     commit-message:
       prefix: chore(example-deps)
-      include: scope
+      # No `include: scope` here: Dependabot appends its own `(deps)` when that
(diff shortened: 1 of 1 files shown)


anilcancakir:
@kodizm review

Issue #29, opened 2025-12-24
chore: Refactor, Document & Test Core Utilities (Typography, Layout, Spacing, Backgrounds)


anilcancakir:
## Description
Following the standard set by `BorderParser` implementation (Feature #12, PR #28), we need to update all remaining core utility parsers to match the new quality standards.

This involves:
1.  **Docs:** Creating Tailwind-style documentation in `docs/` using `<x-preview>` components.
2.  **Examples:** Creating minimal, iframe-ready example pages in `example/lib/pages/`.
3.  **Tests:** Writing feature-based widget integration tests in `test/widgets/` folders.
4.  **Theme:** Ensuring all defaults are exposed in `WindThemeData` and defaults files.

### 1. Typography (TextParser)
- [ ] Docs: `docs/typography.md`
- [ ] Examples: `example/lib/pages/typography/` (color, size, weight, align...)
- [ ] Tests: Move/Update `test/widgets/w_text_test.dart` to `test/widgets/w_text/typography_test.dart`
- [ ] Theme: Check font sizes/weights/colors integration

### 2. Layout & Flexbox (DisplayParser, FlexboxGridParser)
- [ ] Docs: `docs/layout.md` & `docs/flexbox.md`
- [ ] Examples: `example/lib/pages/layout/` (flex, grid, gap, justify...)
- [ ] Tests: `test/widgets/w_div/layout_test.dart`

### 3. Spacing & Sizing (Padding, Margin, Sizing)
- [ ] Docs: `docs/spacing.md` & `docs/sizing.md`
- [ ] Examples: `example/lib/pages/spacing/` & `example/lib/pages/sizing/`
- [ ] Tests: `test/widgets/w_div/spacing_test.dart` & `sizing_test.dart`

### 4. Backgrounds (BackgroundParser)
- [ ] Docs: `docs/backgrounds.md`
- [ ] Examples: `example/lib/pages/backgrounds/` (color, image, gradient...)
- [ ] Tests: `test/widgets/w_div/background_test.dart`

## Acceptance Criteria
- All new docs use `x-preview` and follow the style guide.
- All example pages are added to `routes.dart`.
- All new widget tests pass.
- Test files organized in folders (`test/widgets/w_div/`, `test/widgets/w_text/`).

PR #197, opened 2026-09-07
fix(w_text): capitalize every word, not just the leading letter

anilcancakir:
## What

`capitalize` raised the first letter of the string instead of the first letter of every word. It now walks word initials, matching CSS `text-transform: capitalize`, the doc page and the gallery.

- Whitespace runs survive exactly as typed.
- Leading punctuation is skipped: `"quoted words"` becomes `"Quoted Words"`.
- The rest of each word is left alone: `the HTTP client` becomes `The HTTP Client`.
- An apostrophe does not open a word, so `don't` does not become `Don'T`.
- Each initial goes through the locale mapping from #195, so under `Locale('tr')` `izleyici ışıkları` becomes `İzleyici Işıkları`.

Also adds the Locale-aware Casing section to the gallery's Text Transform page, the demo surface #195 landed without.

## Why

The code was `text[0].toUpperCase() + text.substring(1)`, while `doc/typography/text-transform.md` has read "converts the first character of each word" since the page was written, the gallery's quick reference advertises `Title Case`, and `references/tailwind-divergence.md` carries no `capitalize` row. So the behaviour contradicted every place that describes it, and it was an accidental gap rather than a deliberate Wind divergence.

The gallery is where it hid: the demo row's source string `The quick brown Fox` already starts with a capital, so the `capitalize` row rendered identically to the `normal-case` row beneath it.

**Expect on upgrade:** any multi-word string carrying `capitalize` now renders every word capitalised, where before only the leading letter moved.

## Testing

Five new tests, all proved red before the fix: per-word casing, the acronym left as typed, leading punctuation skipped, whitespace preserved, and the Turkish per-word mapping. The existing `Hello world` assertion flips to `Hello World`.

- `dart analyze`: clean in the package and in `example/`
- `dart format`: no diff
- `flutter test`: 1753 passing, the one pre-existing skip
- `./tool/coverage.sh 90`: 94.8%
- `flutter build web` on the example, then the page loaded in a browser in both light and dark themes

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed the `capitalize` text transform to capitalize the first letter of every word.
  - Preserved acronyms, existing casing, whitespace, and leading punctuation.
  - Added locale-aware capitalization for Turkish and Azerbaijani text.

- **Documentation**
  - Updated text-transform guidance and examples to reflect the corrected behavior.
  - Added locale-aware casing examples to the typography gallery.


anilcancakir:
Both findings taken in 7b81c27.

The digit one was right and it corrected my assumption rather than confirming it. I rendered nine samples in Chromium instead of reading the spec: `4th of july` gives `4th Of July`, `123abc` is untouched, `_underscore lead` gives `_underscore Lead`, and `"quoted words"` still gives `"Quoted Words"`. So the prefix worth skipping is punctuation that separates words, while a digit or a connector character joins the word (UAX #29 Numeric and ExtendNumLet). The class is now `[^\p{L}\p{N}\p{Pc}]`, which covers the underscore case the suggested `[^\p{L}\p{N}]` would still have capitalised. It also killed a claim I had written into the first pass: `3d printing` is `3d Printing`, not `3D Printing`.

The `setUp(WindParser.clearCache)` is in as well. It was the one file in `test/widgets/w_text/` without it.

Gates re-run: analyze clean, format no diff, 1754 tests passing, coverage 94.8%, check-docs 0 issues.

anilcancakir:
Kodizm's three findings are all in now. The first two landed in 7b81c27 (the digit-led word, and the missing `setUp(WindParser.clearCache)`), and the hyphen one landed in 1ecfa3f, where it turned out to be worth fixing rather than documenting.

Measuring it replaced my model instead of patching it. Rendered in Chromium: `well-known issue` is `Well-Known Issue`, `read/write access` is `Read/Write Access`, `u.s.a. today` is `U.S.A. Today`, `a+b=c sum` is `A+B=C Sum`, while `l'orange soup` is `L'orange Soup`, `wind_ui token` is `Wind_ui Token` and `3rd party support` is `3rd Party Support`. So the rule is not "whitespace separates words, skip the punctuation in front of one". It is: raise a letter unless the character before it continues the word, and the four that continue it are a letter, a digit, connector punctuation and an apostrophe.

The pattern is now one lookbehind, `(?<![\p{L}\p{N}\p{Pc}'’])\p{L}`, which reproduces all 23 rendered strings, including six the previous pass got wrong: `Well-known`, `Mother-in-law`, `E-mail`, `Read/write`, `Foo.bar`, `U.s.a.`. So there is no `capitalize` row to add to `tailwind-divergence.md` after all.

Two more tests cover the hyphen, slash and dot boundaries and the apostrophe that is not one. Gates: analyze clean, format no diff, 1756 tests passing, coverage 94.8%, check-docs 0 issues.

anilcancakir:
Right again, and it reproduced exactly as described. Fixed in fd4ee76.

The class listed what continues a word and left out what attaches to it. NFD `naïve` is `n a i U+0308 v e`, so with no `\p{M}` the `v` read as a word initial and rendered `NaïVe`. Every accented sample already in the suite is precomposed, which is why nothing caught it.

Rendered in Chromium to confirm the target rather than infer it: NFD `naïve approach` is `Naïve Approach`, `éclair shop` is `Éclair Shop`, and `co` + U+00AD + `operate now` is `Cooperate Now`, so the format-character arm is real too. The class is now `[^\p{L}\p{N}\p{Pc}\p{M}\p{Cf}'’]` and reproduces all 27 rendered strings.

Both new tests write the code point as a `\u` escape rather than pasting the character, so the source encoding cannot precompose it and pass for the wrong reason.

Gates: analyze clean, format no diff, 1758 tests passing, coverage 94.8%, check-docs 0 issues.

anilcancakir:
Taken in 18f0631, and rendering it settled the part the review flagged as unconfirmed.

The finding is right: the mark and format run was unconditional, so an invisible character standing alone after a space read as a word continuer and ate the next capital. `hello ` + U+200F + `world` came out `Hello world`. The run now has to follow a word character inside the lookbehind, which keeps the NFD and soft-hyphen cases and restores this one.

The suggested pattern does not survive the measurement, though. Chromium renders `co` + U+200B + `operate here` as `CoOperate Here`, so a zero-width space BREAKS a word, while U+200D gives `Ab` and U+00AD gives `Cooperate` and both join. UAX #29 agrees: U+200B is excluded from the Word_Break Format class. So U+200B comes out of the run as well, and the pattern is now `(?<\p{L}\p{N}\p{Pc}'’[\p{M}\p{Cf}])*)\p{L}`, reproducing all 32 rendered strings against 29 for the suggestion as written.

Two more tests: the standalone format character after a space, and the zero-width space that breaks. Both write the code point as an escape.

Gates: analyze clean, format no diff, 1760 tests passing, coverage 94.8%, check-docs 0 issues.

PR #198, opened 2026-09-07
release: 1.5.1

anilcancakir:
## What

Patch release. Two fixes, both in `WText`'s casing transforms.

- **Locale-aware casing (#195).** `uppercase` / `lowercase` / `capitalize` cast under the ambient locale, so a Turkish app renders `İZLEYİCİLER` and `GÜVENLİK` instead of `IZLEYICILER` and `GÜVENLIK`. With no `Localizations` ancestor the transform falls back to Dart's locale-independent casing, which is what every bare widget test gets.
- **Per-word `capitalize` (#197).** It raised the first letter of the string, where the doc page, the demo gallery and CSS all promise the first letter of every word. The word rule was measured in Chromium over 32 strings: `well-known` is `Well-Known`, `read/write` is `Read/Write`, while `3rd party` stays `3rd Party`, `l'orange` stays `L'orange` and decomposed NFD text no longer capitalises mid-word.

**Expect on upgrade:** a multi-word string carrying `capitalize` now capitalises every word, where before only the leading letter moved.

## Version surfaces

The patch set, five files:

| File | Change |
|:-----|:-------|
| `pubspec.yaml` | `version: 1.5.1` |
| `example/pubspec.yaml` | `version: 1.5.1+1` |
| `dartdoc_options.yaml` | source-link tag `blob/1.5.1/` |
| `llms.txt` | `Version 1.5.1 stable` |
| `CHANGELOG.md` | `[Unreleased]` promoted to `[1.5.1] - 2026-09-07`, tag reference added, compare reference retargeted |

`skills/wind-ui/` needs no version move for a patch: the H1, the `description` prefix and the `fluttersdk_wind 1.5.x` marker all still read right, and the skill's own version went to 2.13.2 with the content change in #197.

## Testing

- `dart analyze`: clean in the package and in `example/`
- `dart format --set-exit-if-changed`: no diff
- `flutter test`: 1760 passing, the one pre-existing skip
- `./tool/coverage.sh 90`: 94.8%
- `dart pub publish --dry-run`: the only warning was this bump sitting uncommitted, so it clears with the commit
- Release-notes extraction (`awk` over the `## [1.5.1]` section, the check that fails the `github-release` job on empty notes): returns both entries

## After merge

`git tag 1.5.1 && git push origin 1.5.1` triggers `publish.yml`: validate, pub.dev publish over OIDC, the GitHub Release, and the registry sync job that pushes `skills/wind-ui/` to `fluttersdk/ai`. pub.dev cannot unpublish, only retract, so the tag push is the point of no return.

## Summary by CodeRabbit

- **Release**
  - Published version 1.5.1.
  - Updated package and example application version information to 1.5.1.
  - Updated documentation links and project metadata to reflect the 1.5.1 stable release.
  - Refreshed release comparison references and added the 1.5.1 release link for easier navigation between versions.


anilcancakir:
Taken in fc0c12e, and the precedent is exactly as described: `git show 1c9b9a9 -- example/pubspec.lock` is the single hunk `-version: "1.4.1"` / `+version: "1.5.0"`, and that commit's message calls out why it is there.

Produced by running `flutter pub get` in `example/` rather than editing the line, and the diff came back as the one line with no `source: path` churn from the gitignored `pubspec_overrides.yaml` and the root `pubspec.lock` untouched.

So the patch-release surface is seven files, not the six the checklist has been carrying. Noted for the next one.

The second note is right too and stays out of this PR: the root `pubspec.lock` records `fluttersdk_wind_diagnostics_contracts` as `source: path` because this workspace develops the bridge alongside wind through a gitignored `pubspec_overrides.yaml`. `pubspec.yaml` carries the hosted `^1.1.0` constraint, which is what pub.dev resolves, so publishing is unaffected.

PR #199, opened 2026-09-07
ci(docs): stop failing the weekly link pass on a transient upstream status

anilcancakir:
## What

The weekly `External Links` pass no longer fails on a 429 or a 5xx, and it now keeps a one-day request cache.

```yaml
--cache
--max-cache-age 1d
--accept '100..=103,200..=299,429,500..=504'
```

Plus an `actions/cache` restore step for `.lycheecache`, pinned by SHA like every other action here.

## Why

Run [34130370451](https://github.com/fluttersdk/wind/actions/runs/34130370451) failed with three 504 Gateway Timeouts from github.com, all of them this repo's own pages:

```
[504] https://github.com/fluttersdk/wind/compare/1.5.0...HEAD  (CHANGELOG.md 9:4)
[504] https://github.com/fluttersdk/wind/releases/tag/1.3.0    (CHANGELOG.md 95:4)
[504] https://github.com/fluttersdk/wind/releases/tag/1.4.1    (CHANGELOG.md 35:4)
```

All three load fine in a browser, and neither obvious explanation survives a look at the source:

- **Not a missing token.** `lychee-action`'s `action.yml` defaults its `token` input to `${{ github.token }}` and `entrypoint.sh` exports it as `GITHUB_TOKEN`, so lychee is already authenticated. A token addresses rate limiting anyway, which arrives as 403 or 429, not 504.
- **Not a missing retry.** lychee's `--max-retries` already defaults to 3, with a 1 second wait.

So the job was failing on the one thing it does not exist to catch. It hunts link rot, and a 5xx says the host is up and unhappy rather than that the link is gone. 404 and 410 still fail the job, which is what rot actually looks like.

The cache is the second half: a same-day re-run reuses the answers instead of asking every host again. The one-day age is deliberately shorter than the weekly cadence, so each Monday pass still checks every URL for real and the cache only helps the manual retry after a transient failure.

## Testing

- `actionlint .github/workflows/docs-link-check.yml`: clean
- `zizmor .github/workflows/docs-link-check.yml`: no findings
- Workflow dispatched on this branch to exercise the real external pass, not just the syntax: [run 34150081246](https://github.com/fluttersdk/wind/actions/runs/34150081246)

## Not included

`--accept-timeouts` stays off. A client-side timeout is not what failed here, and swallowing one would hide a host that hangs rather than answers. It is the next knob if a timeout ever shows up.

## Summary by CodeRabbit

* **Chores**
  * Improved external documentation link checks with temporary caching for faster retries.
  * Scheduled checks now revalidate links weekly.
  * Rate-limit and temporary server errors no longer cause checks to fail; other link failures remain reported.


anilcancakir:
Proved the flag rather than trusting it. The dispatched run passes (939 links, 935 OK, 4 excluded, 0 errors), but GitHub answered every request that time, so that run only shows the workflow is well-formed. So I pulled the exact binary CI installs, `lychee 0.24.2`, and drove it against controlled statuses from `httpbingo.org`.

Defaults, which is what run 34130370451 hit:

```
[404] https://httpbingo.org/status/404 | Rejected status code: 404 Not Found
[429] https://httpbingo.org/status/429 | Rejected status code: 429 Too Many Requests
[504] https://httpbingo.org/status/504 | Rejected status code: 504 Gateway Timeout
🔍 4 Total  ✅ 1 OK  🚫 3 Errors
```

With the accept list this PR adds:

```
[404] https://httpbingo.org/status/404 | Rejected status code: 404 Not Found
🔍 4 Total  ✅ 3 OK  🚫 1 Error     (exit 2)
```

So the 504 and the 429 stop failing the job and the 404 still fails it, which is the whole point: rot is still caught.

The cache half checks out too. A first `--cache --max-cache-age 1d` run writes `.lycheecache` and the second run over the same input answers from it, 568ms down to 200ms with the same 2 OK.

anilcancakir:
Both findings verified against the source and taken in caff7e3, and the cache is gone rather than repaired.

`actions/cache` does declare `post-if: "success()"` at the pinned SHA, so the save is skipped on exactly the runs a retry would follow: after this change a 429 or a 5xx no longer fails the job, so a red run means a real 404. The `restore` plus `save` split with `if: always()` would fix that, but the whole external pass costs under four seconds for 939 links (run 34150081246), so there is nothing worth buying back. Dropping the cache retires the two minors and the `.gitignore` note in the same move: `--max-cache-age 1d` was restating lychee's own default, an accepted 429 or 504 would have been replayed out of the cache for the rest of the day, and `.lycheecache` would have needed ignoring for anyone reproducing the pass locally.

The CHANGELOG entry is in under `### Quality`, which is where `CLAUDE.md` files CI infrastructure and where 1.4.1's four workflow-only entries live. It carries the measurement rather than the intent: with lychee 0.24.2's defaults a 404, a 429 and a 504 all fail the run, and with the accept list the 429 and the 504 pass while the 404 still fails it with exit 2.

What is left in the diff is one flag and its comment. Re-verified after the edit: `actionlint` clean, `zizmor` clean, `tool/check-docs.py` 0 issues, and the workflow dispatched again on this branch.

anilcancakir:
Right, and the code was the wrong half. Fixed in c296a51 by widening the range rather than narrowing the sentence: stopping at 504 covers the failure that prompted this and leaves the same class failing one door down, since Cloudflare answers 520, 521, 524 and 530 for an origin that did not respond, and 508 sits outside it too.

Measured on the pinned lychee 0.24.2 over 503, 504, 508, 521, 524, 530 and a 404:

```
--accept '...,429,500..=504'   ->  🔍 7 Total  ✅ 2 OK  🚫 5 Errors   (404, 508, 521, 524, 530)
--accept '...,429,500..=599'   ->  🔍 7 Total  ✅ 6 OK  🚫 1 Error    (404)
```

The 404 is the only link in that list that is actually gone, and it is the only one that still fails the job. The comment and the CHANGELOG entry both carry the reason for the wider bound now.

`actionlint` clean, `zizmor` clean, `tool/check-docs.py` 0 issues, and the workflow dispatched again on this branch.

PR #206, opened 2026-09-15
fix(w-div): stop a rounded overflow clip erasing the border at the corners

anilcancakir:
## What

`overflow-hidden` erased a container's border at its rounded corners. One line of behaviour, one branch, and it reaches every recipe in the ecosystem that clips and draws a border at the same time.

Reported against a consumer app on iOS: the settings list rows and the status-preview card drew a border along their straight runs and nothing through the curves, which reads as a rendering fault rather than as a clip.

## Why it happened

`WDiv` wrapped every clipped subtree in a `ClipRRect` with `Clip.hardEdge`. That takes no anti-aliasing, so it can only cut along whole pixels. On a square clip that is exactly right and it is the cheapest option. On a rounded one it saws the curve into a staircase, and the thing standing on that curve is the border: `BorderSide.strokeAlign` defaults to `strokeAlignInside`, which puts the stroke's OUTER edge exactly on the clip path, so the staircase eats a 1px line rather than the surface behind it.

Flutter's own guidance reads the same way. `hardEdge` is documented as reasonable "if the container is an axis-aligned rectangle or an axis-aligned rounded rectangle with very small corner radii", and `antiAlias` as recommended "when clipping is needed and the shape is not an axis-aligned rectangle". Wind's rounded tokens run from `rounded-sm` (4px) to `rounded-3xl` (32px), so "very small" does not describe them.

## The change

```dart
// lib/src/widgets/w_div.dart
widgetToBuild = ClipRRect(
  borderRadius: resolvedRadius,
  clipBehavior: resolvedRadius == BorderRadius.zero
      ? Clip.hardEdge
      : Clip.antiAlias,
  child: widgetToBuild,
);
```

The zero-radius case keeps `hardEdge` deliberately, and that is the half worth arguing for: anti-aliasing a straight edge buys nothing, and `antiAlias` carries Flutter's documented bleeding-edge artifact where a child painting right up to the boundary shows through. Scoping the change to the shape that needed it is what keeps that cost off every plain rectangle.

## Tests

Two cases in `test/widgets/w_div/overflow_test.dart`, one per branch, and both were mutation-checked rather than merely written:

- forcing `Clip.hardEdge` everywhere turns `a rounded overflow clip is anti-aliased` red and leaves the square case green
- forcing `Clip.antiAlias` everywhere turns `a square overflow clip stays hard-edged` red and leaves the rounded case green

Pinning the square branch is what stops a later "just anti-alias everything" from passing.

## Docs

`doc/layout/overflow.md` claimed `Clip.hardEdge` in prose and again in the token table. Both corrected, with the reason rather than only the value, since the failure mode is not guessable from the token name.

## Verification

`flutter test` green on the full suite (1807 tests) against Flutter 3.47.4. The original report was re-checked on an iPhone 17 simulator: the settings card's border is continuous through all four corners after the change.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed rounded containers with `overflow-hidden` so borders no longer disappear or appear jagged at the corners.
  - Preserved crisp clipping for square containers.

- **Documentation**
  - Expanded guidance on overflow clipping, rounded corners, and border rendering behavior.

- **Tests**
  - Added coverage for rounded and square overflow clipping scenarios.


anilcancakir:
Both points addressed.

**Major (formatting)** was already fixed in `59e5622`, pushed before this review landed; `Lint & Test` is green on the branch now. `dart format --output=none --set-exit-if-changed .` reports `391 files (0 changed)`.

**Minor (the example gallery)** is fixed in the commit above. Good catch: the gallery row and the doc table are the same claim on two surfaces, and only the doc had moved.

On the scoping note, I verified the reasoning rather than taking the conclusion. `w_image.dart:150` and `w_select.dart:790` build their `ClipRRect` with **no** `clipBehavior` argument, and Flutter's default for that parameter is already `Clip.antiAlias`. So they were never affected: the defect was specific to `w_div` passing `Clip.hardEdge` explicitly, which is what made it the one site to change.

anilcancakir:
@kodizm

PR #207, opened 2026-09-15
feat(w-select): tell the caller when the menu opens

anilcancakir:
## What

`WSelect` offers a caller-driven pagination API (`onLoadMore` + `hasMore`) and then silently resets the list that API is paginating, without telling the caller.

`_toggleMenu` on open does:

```dart
_searchQuery = '';
_filteredOptions = widget.options;
```

That is correct for the widget. It is invisible to the caller, whose cursor survives the reset, so the next scroll to the bottom asks for the page **after** the one the reader can no longer see.

## The reproduction

A searchable timezone select over a paged endpoint, reported on iOS:

1. open the menu, 20 rows
2. scroll to the bottom, page 2 loads, 40 rows
3. close
4. reopen, back to 20 rows (the widget reset), but the caller's cursor still says 2
5. scroll to the bottom, the caller asks for page **3**

Page 2's twenty identifiers are now unreachable without searching for them, which is the same symptom the pagination API was added to remove.

The searched case is worse: reopening clears `_searchQuery` and shows the unfiltered list, while the caller still holds the old query, so `onLoadMore` appends page 2 **of that query** onto an unfiltered list, and `hasMore` may be false for a query nobody is running any more.

## The change

`onOpen`, on `WSelect` and forwarded through `WFormSelect`.

A callback rather than internal state, deliberately: the cursor belongs to whoever owns `onLoadMore`, and nothing inside this widget needs it. Resetting it here would mean the widget owning half of a caller's state.

A select that passes no `onOpen` behaves exactly as before, which the third test pins rather than leaves implied.

## Tests

Three cases in `test/widgets/w_select_on_open_test.dart`: it fires on open, it fires **again on a reopen** (the case it exists for, and the one a naive "fires once" test would miss), and a select without the callback still opens cleanly.

## Gates

`dart format` 392 files 0 changed, `flutter analyze` clean, `flutter test` 1808 passed against Flutter 3.47.4.

## Why this is a separate PR from #206

#206 is the rounded-clip fix and shares no file with this. Keeping them apart means neither waits on the other's review.

## Downstream

`magic_starter`'s timezone select is the caller that hit this; its fix consumes `onOpen` and is held on its own PR until this releases, so the rest of that work is not blocked behind a wind release.

## Summary by CodeRabbit

* **New Features**
  * Added an optional `onOpen` callback to select controls, including single- and multi-select form controls.
  * The callback runs whenever the dropdown opens, enabling paginated applications to refresh or restore loading state.

* **Bug Fixes**
  * Prevented outdated asynchronous search and pagination results from appearing after reopening a menu or changing its options.
  * Fixed reopened menus remaining stuck on a loading spinner.
  * Preserved rounded border corners when using clipped containers.

* **Documentation**
  * Added guidance for using `onOpen` with searchable, paginated selects.


anilcancakir:
Both addressed in `9539ae4`.

**Major, the sync surfaces.** Correct, and your note that CI cannot catch it is the part worth repeating: `check-docs.py` verifies links and previews, not prop completeness, so it passed while the prop was missing from two places. Now carried by `doc/widgets/w-select.md` (constructor signature and Props table), `skills/wind-ui/references/widgets.md` (the `WSelect` signature, the behaviour bullet, and BOTH `WFormSelect` passthrough lists, there are two), and the skill version plus its stamp move with them.

**Minor, the doc comment overreaching.** Good catch, and it is the more interesting of the two. The advice is only safe for the caller shape in the report, where `options` stays at page one because `onLoadMore` returns rows instead of mirroring them. A caller that mirrors fetched pages into `options` already holds all forty on reopen, and resetting its cursor as instructed would re-fetch page two and append duplicates, which is a worse bug than the one this closes.

Stated in both places a reader might stop at, since the callback name reads as unconditional:

> Reset the cursor only if `options` resets too.

Thank you also for checking the two things that could have been wrong and were not: `didUpdateWidget` re-seeding when a caller shrinks `options` from inside `onOpen`, and the `setState` placement matching `WPopover`. Both were assumptions I had not verified.

Gates: `dart format` 0 changed, `flutter analyze` clean, `check-docs.py` 0 issues, select suites green.

anilcancakir:
Fixed in `7f5c71a`, and I took your suggested direction rather than the cheaper one.

You are right about the cause and it was self-inflicted: the previous commit replaced the passthrough comment with a blanket substitution that matched BOTH lists, and only one of them is `WFormSelect`. So the skill documented `onOpen` on `WFormMultiSelect` while the widget's constructor stopped at `hasMore`. A reference that makes an agent write a compile error is worse than one that says nothing.

**Forwarded rather than removed from the list**, for the reason you gave: a multi-select stays open while the reader picks, so a paginating caller meets the open-reset more often there than on a single select. It is now declared on the public constructor, the inner one, both field sites and the passthrough, matching `WFormSelect`.

There is now a test per wrapper, `reaches a WFormMultiSelect too` and `reaches a WFormSelect too`. Your closing observation is exactly why: nothing covered the skill's claims, which is how the line got through. A passthrough that is documented and not wired now fails a test instead of misleading a reader.

On `dart format`: your caution was right to record, and for what it is worth the repo-pinned run here is clean, `392 files (0 changed)`.

Gates: `dart format` 0 changed, `flutter analyze` clean, `check-docs.py` 0 issues, 5 cases in the onOpen suite green.

anilcancakir:
Rebased onto master now that #206 has merged, so this branch carries the clip fix too and the `[Unreleased]` section holds both entries (Added then Fixed, matching the file's order elsewhere). No behaviour moved in the rebase; the only conflict was the changelog and both sides are kept.

Re-verified on the rebased head: `dart format` 392 files 0 changed, `flutter analyze` clean, `check-docs.py` 0 issues, `flutter test` 1812 passed.

anilcancakir:
One more commit on the branch, from a defect that fell out of a downstream review rather than from yours.

`202a27b` - **reopening a `WSelect` strands the menu on a spinner when a search was in flight.** The open branch this PR already touches resets `_searchQuery` and `_filteredOptions`, but `_isSearching` survived it, and the only place that lowers that flag is `_filterOptions`'s resolve path at `w_select.dart:485`, guarded on `_searchQuery == query`. The reset guarantees that comparison fails, so the response that eventually lands clears nothing: the reopened menu shows `loadingBuilder` over an empty list, with no rows, no empty state, and no recovery short of another keystroke.

It is not exotic. A debounced caller makes the window the debounce rather than the network: type, close inside 300ms, reopen. That is exactly the shape `magic_starter`'s timezone select has, and it is where this surfaced.

Reproducer in `test/widgets/w_select_on_open_test.dart`, under a new `WSelect reopen` group. Before the patch it fails on `pumpAndSettle timed out`, because a spinner never settles, so the assertions are explicit pumps and `findsNothing` on `CircularProgressIndicator`. The late response is also asserted not to resurrect the spinner or overwrite the restored list. Mutation-verified: deleting the one added line reports `Expected: no matching candidates / Actual: Found 1 widget with type "CircularProgressIndicator"`.

The zero-radius / `WFormMultiSelect` findings from the previous rounds stand as answered.

`dart format` 218/0, `flutter analyze` clean, full suite `+1813` (was 1812).

anilcancakir:
Confirmed both halves of the Minor, including the one you labelled pre-existing, and closed the class rather than the instance. `8f713bf`.

**The empty-query case.** Reproduced exactly as described: type, delete back to empty, close and reopen inside the window, and the late response passes `_searchQuery == query` because `''` is equal to itself across the reset. It overwrites the `_filteredOptions = widget.options` the reopen restored.

**The `_loadMore` case is real too**, and it took a second measurement to see. My first reproducer asserted on `maxScrollExtent` and passed while the bug was present: the reopen recomputes `_openUpward`, so the viewport height moves by 20px and the extent is not a row count across a reopen. `find.text` on an appended row is vacuous for the opposite reason, since the menu is a lazy `ListView` and a row stitched onto the end is never built. Walking the reopened list to its end settles it: with 30 options, a page of `B0..B9` held open across the reopen, `find.text('B9')` reports **1** before the patch. So the page does append onto the restored list.

**One mechanism for both, your suggestion.** The guards were asking the wrong question. It is not "is this the answer I last requested" but "is the list I was answering still on screen", which a query string cannot express and which `_loadMore` had nothing at all to express. So `_listEpoch` is bumped in the open branch, both paths capture it before their await, and a response whose epoch has moved is dropped. `_loadMore` still lowers `_isLoadingMore` when it drops a page, because leaving it raised would refuse every later scroll.

Two new cases, one per path, each mutation-verified separately: deleting the `if (epoch == _listEpoch)` in `_loadMore` fails `the stale page appended onto the restored list`, and deleting the `epoch == _listEpoch` term in `_filterOptions` fails `the stale empty-query response wrote over the restored list`.

`dart format` 218/0, `flutter analyze` clean, full suite `+1815` (was 1813).

anilcancakir:
@kodizm please re-review. The check on `8f713bf` reported `Not reviewed`, so the `_listEpoch` change and its two new tests have not been looked at yet.

anilcancakir:
Confirmed and fixed in `be9fef5`. You are right that it is pre-existing, and you are also right that it makes my own doc comment false: it said "bumped every time the menu resets its own list" while covering only the open path.

Reproduced as described: menu open, page two in flight, caller swaps `options`, and the stale page appends onto the refreshed list. `find.text('B9')` finds it after walking the refreshed list to its end. One `_listEpoch++` next to the assignment in `didUpdateWidget`, and the comment now says "every open and every caller-driven change of `options`", which is the whole invariant rather than half of it.

Mutation-verified by deleting just that line.

Three places replace the list wholesale and all three are now inside the epoch. That is the count I should have taken before writing the invariant rather than after you found the third.

One note on the reproducer, since it is the third time in this PR: `pumpAndSettle` after the `options` swap times out, because the load-more spinner is up at that point and a spinner never settles. Counted pumps again.

`dart format` 218/0, `flutter analyze` clean, full suite `+1816` (was 1815), this file 9 cases.

anilcancakir:
@kodizm please re-review `be9fef5`. The check reported `Not reviewed` again; that commit is the `didUpdateWidget` epoch bump you raised last round, plus its reproducer.

anilcancakir:
Confirmed, and this one is mine end to end: `be9fef5` reopened the strand through the door it had just closed. Fixed in `f9b8ef0`.

Reproduced exactly as you traced it. The epoch bump drops the stale search response, which is right, but the resolve path is the only place that lowers `_isSearching` and it is guarded on the epoch, so the drop takes the flag down with it. The empty query is what makes it reachable: `didUpdateWidget` re-runs the search only when the query is non-empty, so a search for `''` has nothing behind it to raise and lower the flag again. Test reports `Found 1 widget with type "CircularProgressIndicator"` before the fix.

`_isSearching = false` beside the bump, as you said, mirroring the open branch. Mutation-verified by deleting just that line.

**The rule the epoch implies and that I had not stated**: every place that replaces the list has to lower the in-flight flag, because after the reset nobody else owns it. The open branch already did it, `_loadMore` does the same thing a different way by lowering outside the epoch check, and `didUpdateWidget` was the one site that bumped without it. Three reset sites, three flags, and I added the third guard without asking what the first two also do. That is the same mistake shape as the round before, where I wrote the invariant before counting the sites.

`dart format` 218/0, `flutter analyze` clean, full suite `+1817`, this file 10 cases.

anilcancakir:
@kodizm please re-review `f9b8ef0`. The check reported `Not reviewed` again; that commit is the `_isSearching = false` beside the `didUpdateWidget` epoch bump, the Major you raised last round, plus its reproducer.

anilcancakir:
Taken, in `bc31070`. You called it cosmetic and non-blocking and I agree with both, but it is the last place in the file that lowered a flag it might not own, and that is the rule the previous commit wrote down: leaving one exception to it is how the rule stops being one.

Reproduced as described: type `p`, type `pa`, fail A, and the spinner drops while B is still out, leaving the pre-search list on screen looking settled. The catch now carries the same two terms as the success path. Nothing is stranded by it, since a newer request owns the flag and lowers it, and a reset that moved the epoch has already lowered it itself.

Mutation-verified: removing just the catch's guard turns the new case red.

That makes four reset-or-resolve sites in this widget and all four now agree on who owns the flag. Counting them first would have saved three rounds, which is the actual lesson from this PR.

`dart format` 218/0, `flutter analyze` clean, full suite `+1818`, this file 11 cases.

anilcancakir:
@kodizm please re-review `bc31070`. The check reported `Not reviewed`. That commit takes the non-blocking Minor from your last pass: the `catch` in `_filterOptions` now carries the same two guard terms as the success path, with a reproducer and a mutation check. Nothing else moved.

PR #208, opened 2026-09-16
fix(w-input): clear the keyboard AND its toolbar, not just the caret

anilcancakir:
A focused `WInput` came out from under the keyboard and straight under the keyboard toolbar. Reported off a TestFlight build against a three-line update composer: tapping it left half of its first line showing and the reader scrolled by hand.

## Two occluders, one reported

`EditableText` scrolls the CARET into view with `scrollPadding` around it, on focus and on every metrics frame. On one line the caret and the field are the same box and that is right. On several they are not: the caret sits on line one, so line one clears the keyboard and everything below it stays under it. `scrollPadding` is now widened below by the field's measured height, which points Flutter's own mechanism at the whole field.

**The second occluder is the toolbar, and nothing reports it.** `WKeyboardActions` draws its bar in an overlay at `bottom: viewInsets.bottom`, so it sits ON TOP of the keyboard; the engine reports the keyboard and knows nothing about the bar. `WKeyboardActions` now measures its own bar and publishes the height through a new `WKeyboardToolbarInset`, and `WInput` adds it. The bar is iOS-gated and widget tests run as Android, which is why none of this reproduced in a test until one opted into iOS.

The height is measured rather than derived from `maxLines` or a constant: a line's height depends on the resolved text style and the field's padding, and the toolbar's on the ambient `IconButton` theme and a consumer's `toolbarClassName`.

## What I got wrong first, and how the harness hid it

The first version added a `Scrollable.ensureVisible` instead. A review pushed back and the measurements are the useful part:

- **It did nothing.** `TestFlutterView.viewInsets`'s setter already calls `onMetricsChanged` (`flutter_test/lib/src/window.dart:1207`), and my harness called `handleMetricsChanged()` after it. `EditableText` de-dupes its own post-frame scroll and this widget did not, so wind got two callbacks to Flutter's one and ran last. With a single dispatch the field lands byte-identically with and without it.
- **It was harmful.** The harness put the field at the page bottom, so every scroll ended at `maxScrollExtent` and the assertion passed on the page running out rather than on where the field was aimed: 584 at every field height, `pixels == maxScrollExtent` each time. With 900px below the field, `keepVisibleAtEnd` scrolled it **308 pixels off the top of the screen**.

So the harness is fixed here too: one dispatch, 900px below the field, the field positioned just above the fold rather than dragged past, and exact expected values instead of `lessThanOrEqualTo(600)`, which passed throughout the version that scrolled the field off the screen.

## Verification

5 tests in `test/widgets/w_input_scroll_into_view_test.dart`, covering a multiline field, a single-line one, a field already in view, the padding growing with the field, and the iOS toolbar being published and consumed.

Mutation-verified per mechanism: flattening `scrollPadding` back to `EdgeInsets.all(20)` fails three assertions, and dropping the `WKeyboardToolbarInset` term fails the toolbar one.

Measured on a 900px viewport with a 300px keyboard, field bottoms: 550 at 1, 3, 6 and 14 lines, with the top on screen at every height.

`dart format` 219/0, `flutter analyze` clean, full suite `+1823`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved keyboard avoidance for multiline and single-line input fields.
  - Focused fields now scroll fully into view above the software keyboard, including the iOS keyboard toolbar.
  - Scroll positioning follows the caret as users type.
  - Scroll positioning updates correctly when keyboard or screen metrics change.
  - Fields already visible remain in place instead of being unnecessarily repositioned.
  - Tall fields keep the active caret visible while allowing content above it to extend beyond the viewport.

- **Documentation**
  - Added guidance on keyboard toolbar occlusion and custom content positioning.


anilcancakir:
Confirmed and fixed in `fa57451`. The Major is exactly right and I had shipped the failure I claimed to have removed.

**Reproduced first.** With the caret put at the end of the text, a 156px field came to rest with its bottom **112px** clear of the keyboard where line-one gives 42. Mutation-verified after the fix by restoring the full-height term: a 40-line field reports `top: -268.0`, so the top really does leave the screen, which is the same failure the removed `ensureVisible` had.

The term is now how much of the field sits BELOW the caret, read from the render editable's own caret rect. It shrinks to nothing as the caret descends, so a field taller than the visible area degrades to caret visible with the overflow below it rather than to a field yanked upward. Measured with the caret at the end: tops at 416, 374 and 262 for 10, 25 and 40 lines, all on screen.

New test `a field taller than the viewport keeps its top on screen` walks those three heights with the caret at the end. The two exact-value assertions moved 550 to 564, which is the caret-relative arithmetic rather than a regression.

**The Minors, all three.** The dartdocs on `_measureForScrollPadding` and `_scrollPadding` still described the removed `ensureVisible`, its deferral and `alignmentPolicy`; rewritten to say what the method does. The CHANGELOG entry said the same thing and now describes `scrollPadding`. `debugDefaultTargetPlatformOverride` is in a `try` / `finally`. And `WKeyboardToolbarInset` is documented: a new section in `doc/widgets/w-keyboard-actions.md` with a ToC entry, a paragraph in `skills/wind-ui/references/forms.md`, and the SKILL stamp at `2.17.0`.

**Not covered, and I am saying so rather than claiming otherwise.** The auto-grow case you named: `_measuredHeight` refreshes only on focus and on a metrics change, so a field growing from `minLines: 3` toward `maxLines: 6` as the reader types keeps the height it was measured at. The caret term moves with every rebuild, so the reserve stays roughly right, but the height half is stale until the keyboard next moves. I did not want to add a third mechanism to this getter in the same round.

`tool/check-docs.py` 0 issues, `flutter analyze` clean, `dart format` 219/0, full suite `+1824`.

anilcancakir:
Confirmed and fixed in `c7bba30`. Your diagnosis was exact, including why my test hid it: setting the selection before the tap hands the one build that counts a caret already at the end.

**Reproduced first.** Tap a 40-line field at the top, raise the keyboard, then type downwards: the padding sat at its focus-time **582 through forty typed lines**.

Three things were needed and I found each by measuring rather than reasoning, which is worth recording because the first two both looked sufficient:

1. **A controller listener.** As you say, nothing else rebuilds this widget on a text change.
2. **The recomputation deferred to a post-frame.** The controller notifies SYNCHRONOUSLY and the render editable has not laid the new text out yet, so reading the caret rect then returns the one from before the keystroke, the difference is zero, and the listener bails for good. With the listener alone the padding still read 582.
3. **A `bringIntoView` afterwards.** The reveal that came with the keystroke already used the stale padding, so a corrected padding does not move the view back on its own. With 1 and 2 the padding fell 582 to 36 exactly as it should and the field stayed 534 pixels above the top of the screen.

**Your Major also exposed a wrong assertion in the test I added last round.** With the caret on the last line the caret sits at the field's BOTTOM, so a field taller than the visible area cannot both show its caret and keep its top on screen: overflowing upward is the correct degradation, and `top >= 0` was demanding the impossible. It passed only because the padding was wrong in a way that hid it. It now asserts the bottom, which measures 572 at 10, 25 and 40 lines.

**Two things I am flagging rather than hiding.**

The typing test is in its own file. It passes alone and fails after its neighbours in one file, by about 400 pixels, for state I could not identify and did not want to paper over with a bigger tolerance.

It asserts the PADDING rather than geometry, and that is deliberate: a 6-line field stays on screen either way, so the geometric version passed with the listener removed. The padding assertion dies under that mutation (`Actual: 106.0` against a bound of 100).

Both Minors done: the doc section is now outside the `styling-examples` anchor with its own, and the missing word is back.

`tool/check-docs.py` 0 issues, `flutter analyze` clean, `dart format` 219/0, full suite `+1825`.

anilcancakir:
All three taken, and you were right to re-check the one I said was done.

**The CHANGELOG.** I replaced the paragraph beside it last round and reported the whole thing fixed, which was wrong: the `keepVisibleAtEnd` / `ensureVisible` sentence was still there describing code that is not in the branch, and nothing described the listener, the deferred recompute or the `bringIntoView`. Rewritten. `grep -rn "keepVisibleAtEnd\|ensureVisible" CHANGELOG.md` is now empty.

**The mutating getter.** Fair, and worse than it looks: rearming the difference test does not only lose a comparison, it swallows the corrective `bringIntoView` that follows it, so a stray read would have brought back the exact defect this commit fixes. The clearance is recorded once at the top of `build` now and the getter only reads.

**The ordering shift, which you could not reproduce and I now cannot either.** You are right, and the reason is in this PR: the tall-field test loops three heights on one tester, and until I added a per-iteration `tester.view.reset()` it left state that moved the typing test by 400 pixels. That reset landed in the same commit as the separate file, so the file was already unnecessary when I wrote the comment saying I had not identified the cause. I merged the test back, ran the file, `+7` green, and deleted `w_input_typing_scroll_test.dart`. The comment now names the polluter instead of shrugging at it.

Mutation re-verified after consolidating, since the tests moved: removing `_controller.addListener(_onCaretMoved)` fails two assertions (`740.0` and `106.0`).

`flutter analyze` clean, `dart format` 219/0, full suite `+1825`, `tool/check-docs.py` 0 issues.

anilcancakir:
Thanks. One thing left after your pass, from CI rather than from review: `codecov/patch` was 95.23% against a 95.34% target, three lines short.

All three were in `WKeyboardActions` and one of them was a real gap: nothing covered the toolbar's teardown, so a published height that outlived its bar would have had every field below it reserving room for something not on screen. `1e9f151` adds that case (unfocus, inset back to zero) and patch coverage goes to 96.83%.

The two still uncovered are the no-index build path and the bounded layout retry. Neither is reachable without a toolbar that refuses to lay out, and I would rather leave them uncovered than reach for a suppression.

anilcancakir:
Taken in `a4be9b1`: the teardown block now sits below the `padding >= toolbar` assertion, so each assertion is beside the state it describes. You are right that it was correct only by accident of the capture order, which is exactly the kind of thing that stops being true when someone later re-reads one of those values in place.

PR #135, opened 2026-06-30
fix(layout): treat w-full on a Row child as flex-1 (#122)

anilcancakir:
## Summary

Fixes #122. A child carrying `w-full` inside a `flex flex-row` aborted layout with `RenderBox was not laid out`: a Row hands non-flex children an unbounded main-axis constraint, and the child's `SizedBox(width: double.infinity)` cannot resolve there. (A `w-full` footer button silently blanked an entire BottomSheet in a real app.)

## Fix

The Row composer (`_composeFlex`) now wraps a **bare** `w-full` direct child in `Expanded` (and defaults the row to `MainAxisSize.max`). `Expanded` hands the child a bounded width share, inside which its existing infinite-width box resolves, so `w-full` fills the row exactly like `flex-1`.

Design notes:
- The fix lives entirely in the **parent** composer, so nested (non-direct) `w-full` children keep their bounded `SizedBox` behavior and are never turned into an `Expanded` outside a Flex (no regression for `Row > WDiv(w-32) > WDiv(w-full)`).
- A prefixed `md:w-full` is intentionally **not** auto-expanded: it is conditional, and unconditional `Expanded` would break the responsive intent. Use `md:flex-1`.
- `w-full` combined with `flex-1` defers to the existing `flex-1` self-wrap (no double-wrap).
- `flex-1` remains the idiomatic, recommended choice.

## Tests

`test/flex/wfull_in_row_test.dart`: fills the row (no exception, wrapped in `Expanded`), coexists with a fixed-width sibling, nested `w-full` stays bounded (regression guard), `flex-1 w-full` does not double-wrap, column `w-full` unaffected.

## Definition of done

- `dart analyze`: clean. `dart format`: no diff.
- `flutter test`: 1557 pass (1 pre-existing skip), coverage 91.4% ≥ 90%. Existing `test/flex` + `test/widgets/w_div` suites stay green.
- Post-change sync: `doc/layout/flexbox.md` note, skill (Core Law 4 + §5/§6/§12/community + `references/layouts.md` + `tailwind-divergence.md` reconciled, version 2.8.0), `CHANGELOG.md`. README not touched.

> Note: `SKILL.md` 2.8.0 bump overlaps with the sibling PRs; last to merge takes a trivial one-line version conflict (keep 2.8.0).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a layout error when using `w-full` as a direct child inside a horizontal flex row.
[...]

PR #209, opened 2026-09-16
release: 1.6.0

anilcancakir:
## What

Minor release. It carries the three PRs merged since 1.5.3, one placement fix found reviewing them, and a README rewritten for the page pub.dev actually ranks.

- **`WSelect.onOpen`, plus both form wrappers (#207).** Opening the menu clears the search and restores `options`, which is right for the widget and invisible to the caller that owns `onLoadMore` and `hasMore`. Its cursor survived that reset, so the next scroll asked for the page after the one the reader could no longer see. A callback, because the cursor belongs to whoever owns the pagination and nothing in this widget needs it. This is the release `magic_starter` is holding its timezone select on.
- **Two `WSelect` async holes closed with a list epoch (#207).** A response in flight across a reopen no longer writes to the list the reopen restored, and a search closed before its response landed no longer strands the reopened menu on a spinner.
- **A focused `WInput` clears the keyboard AND the toolbar above it (#208).** `EditableText` reveals the caret rect, which is the field only while the field is one line tall. `WInput` now widens `scrollPadding` by whatever sits below the caret and adds the `WKeyboardActions` bar height, which the engine reports nowhere.
- **A rounded `overflow-hidden` no longer erases the border at the corners (#206).** `Clip.hardEdge` cannot anti-alias, so it sawed the curve into a staircase and ate the 1px stroke standing on it. Rounded clips are `antiAlias` now; the zero-radius branch keeps `hardEdge` and a test pins both.

Semver: `onOpen` is an additive optional parameter, so this is a minor and not a patch. `magic_starter` currently declares `fluttersdk_wind: ^1.5.4`, a version that was a prediction rather than a release; that floor moves to `^1.6.0` in its own PR.

## The fix found while preparing the release

`WSelect` fired `onOpen` INSIDE the `setState` closure that resets the list, ahead of the post-frame callback that mounts the overlay. A caller throwing there abandoned the closure half done: `_isOpen` true, `_overlayController.show()` never scheduled, so the menu was marked open with nothing on screen. Every other caller callback in the file already fires outside `setState`, and this one now does too (`cd28935`).

## Post-change sync the merged PRs left behind

`CLAUDE.md` asks for five surfaces on every code change, and two were short:

- `example/lib/pages/forms/select_pagination.dart` still demonstrated the defect `onOpen` exists to remove: it kept its cursor across a reopen. It now resets through `onOpen`, which also makes the demo the worked example for the new parameter.
- `doc/widgets/w-input.md` never mentioned the keyboard clearance. The behaviour was documented only on the `WKeyboardActions` page, which is where a reader looks last. New `Keyboard Clearance` section, ToC entry included.

## README

The comparison table of six other packages is gone. It ranked alternatives and spent a paragraph each on what Wind is not, on the one page pub.dev both ranks as a search field and renders as the landing page.

What replaces it: the positioning sentence and a runnable example inside the first forty lines, the raw Flutter equivalent folded into a `<details>` block so the verbosity argument is made by the diff, a capability list that reaches the recent work (keyboard clearance, activation from a keyboard, a gamepad and a remote), and a `className` rules table an agent can act on without opening the docs site. The ecosystem gets one footer sentence instead of a table, which is where flutter_bloc, firebase_auth, forui and TanStack Query all keep it. `dart.dev`'s package page guidance covers the rest: no repeated package name at the top, visual content early, one copy-pasteable example, and terms repeated because readers use in-page search.

## The six stamps

| File | Change |
|---|---|
| `pubspec.yaml` | `1.5.3` to `1.6.0` |
| `example/pubspec.yaml` + `example/pubspec.lock` | `1.5.3+1` to `1.6.0+1`, lockfile tracked via `flutter pub get` |
| `dartdoc_options.yaml` | source link tag to `1.6.0` |
| `llms.txt` | `Version 1.5.3 stable` to `1.6.0` |
| `skills/wind-ui/SKILL.md` | description, body stamp and title from the 1.5 line to 1.6 (skill `version:` stays at `2.17.0`, it was bumped with the API in #207) |
| `CHANGELOG.md` | `## [Unreleased]` to `## [1.6.0] - 2026-09-16`, fresh `[Unreleased]`, link references |

The promotion also repairs what the merged PRs left: a `### Fixed` heading with no blank line under it, three entries naming only their `lib/` files where the convention lists doc, example and skill paths, and five entries with no PR number where the file uses them 22 times.

## Gates

- `dart analyze`: no issues
- `dart format --output=none --set-exit-if-changed .`: 393 files, 0 changed
- `flutter test`: **1825 passed**, 1 pre-existing skip
- `./tool/coverage.sh 90`: **95.4%**
- `python3 tool/check-docs.py`: 72 doc pages + 3 linking files against 171 demo routes, 0 issues
- `dart pub publish --dry-run`: 0 warnings, 1 hint (the local `pubspec_overrides.yaml`, which CI does not have)

## After this merges

`git tag 1.6.0 && git push origin 1.6.0`, which triggers `publish.yml`. Two packages are waiting on it: `magic_starter` #135 cannot resolve from pub.dev until 1.6.0 is up, and its floor becomes `^1.6.0` rather than the `^1.5.4` it predicted.

Author: Anılcan Çakır <anilcan.cakir@gmail.com>

## Summary by CodeRabbit

* **New Features**
  * Added an `onOpen` callback for select menus, including form-based selects.
  * Select menus can reset search and pagination state when reopened.

* **Bug Fixes**
  * Improved keyboard clearance for focused multiline inputs.
  * Prevented stale search results and loading states when reopening select menus.
  * Preserved rounded borders when content is clipped.

* **Documentation**
  * Reworked the README with updated installation, usage, features, and platform guidance.
  * Added documentation for input keyboard clearance.

* **Chores**
  * Updated the package and example application to version 1.6.0.


anilcancakir:
Both Majors confirmed and fixed in `91e8474`, along with the Minor. You were right on all three, and the two Majors are the same mistake twice: I wrote a demo that reproduces the bug the parameter exists to remove.

**Major 1, the cursor derived from `_users.length`.** Reproduced exactly as you described. The cause is upstream of the arithmetic: `_onLoadMore` was appending SEARCH results into `_users`, so the list a reopen restores was not the unfiltered list at all. The loader now RETURNS its rows and pushes nothing, which is what the changelog entry already says a paginating caller should do and what `magic_starter` does: `WSelect` owns the visible list once a search has filtered it and rebuilds it whenever `options` changes identity, so appending from the caller throws the filtered list away mid-scroll. With `_users` staying the first page, the reopen cursor is page one and no page is skipped.

**Major 2, `_hasMore` written without `setState`.** Confirmed against `w_select.dart:363`, which reads `widget.hasMore`. Both writes in `_onMenuOpen` are inside a `setState` now, and so is the pair in `_onSearch`, which had the same hole for the same reason.

**Minor, the untested `setState` move.** Fair, and the README line you quoted is mine. `test/widgets/w_select_on_open_test.dart` now has `a caller that throws still gets the menu it opened`: it throws from `onOpen`, asserts `takeException()` is the caller's `StateError`, and asserts the option is on screen rather than only `_isOpen` being true. Verified red by putting the call back inside the `setState` ahead of the post-frame mount, green on this branch.

Gates after the change: `flutter test` **1826 passed** with the one pre-existing skip, `dart analyze` clean, `dart format` 393 files 0 changed.

anilcancakir:
@kodizm re-review please: head moved to `91e8474`, which is your two Majors plus the Minor.

- The demo loader no longer pushes into `_users`, so the list a reopen restores is the unfiltered first page and the cursor goes back to it. That closes the skipped-rows repro you walked through.
- `_onMenuOpen` and `_onSearch` both write through `setState` now, so `widget.hasMore` actually moves.
- `w_select_on_open_test.dart` gains `a caller that throws still gets the menu it opened`, verified red with the call back inside the `setState`.

Nothing else moved: no version stamp, no README, no lib change beyond the one in `cd28935` you already cleared.

anilcancakir:
Minor fixed in `1c77bfb`, and wider than you flagged: all four async paths in the page now guard on `mounted`, including `_loadInitialUsers` (fires from `initState`) and `_onCreateTag`, which had the same shape on the tagging select. Half a second of simulated latency is long enough for a reader to leave the gallery, and the gallery is exactly where someone clicks through pages quickly.

Thanks for the four rounds on this one. The two Majors were both mine, from a fix written against the parameter's description rather than against the demo's own state.

anilcancakir:
@coderabbitai four findings, one acted on, two already fixed before the review landed, one refuted.

**1. One-line changelog bullets.** You are right that `CLAUDE.md:54` says it and right that these entries do not. I disagree about which side is wrong, so `af5c8b9` changes the rule rather than the entries. The last three releases all wrote long-form entries, the sibling packages in this ecosystem write them, and the reason is specific: a defect whose symptom is geometry, a race or a stale cursor cannot be reconstructed from its diff, so the entry is where the next reader learns what it was measured against and which alternative was rejected. "One line where the change genuinely is one line" stays.

**2. Rebuild after changing `hasMore`.** Already fixed in `91e8474`, which landed before this review. `_onMenuOpen` and `_onSearch` both write through `setState` now.

**3. Keep search pages out of `_users`.** Same commit, and taken further than suggested: `_onLoadMore` appends to `_users` in NO case and returns its rows instead. `WSelect` owns the visible list once a search has filtered it and rebuilds it whenever `options` changes identity, so appending from the caller even on the empty query would throw the filtered list away mid-scroll. `_users` stays the unfiltered first page, which is exactly what a reopen restores.

**4. Bump the skill frontmatter for `onOpen`.** Refuted, the bump already happened in the PR that added the API: `85a78b5` moved `version:` from `2.15.0` to `2.16.0` and `02e0fba` moved it to `2.17.0`. This release PR carries no API change of its own, so bumping again would claim a surface move that did not happen. What it does move is the `1.5` series naming in the description, the body stamp and the title, which is the release stamp rather than the skill version.

anilcancakir:
Minor fixed in `fe8c467`, and the whole drift with it: `diff CLAUDE.md .github/copilot-instructions.md` is now empty.

You were right that the changelog line was about to become the seventh divergence and the one that matters most, since a Copilot session would have written the shape the rule above it just rejected. The other five were rot with a shelf life: the README acceptance line named 22 widgets and 19 parsers against the real 27 and 20, the docs acceptance line described a check that predates `tool/check-docs.py`, the commands table was missing that script, and the branching section was missing both the branch-lifetime paragraph and the two required status checks.

Merging after this: the tag is what publishes, and `magic_starter` #135 is holding on 1.6.0 being on pub.dev.

PR #210, opened 2026-09-16
fix(theme): scaffoldBackgroundColor disagreed with the app's own canvas

anilcancakir:
Found while reviewing `fluttersdk/magic#164`, which is the PR that makes this matter. It is a latent disagreement that has been here for as long as both halves have, and it becomes a visible defect in the next magic release.

## The disagreement

`scaffoldBackgroundColor` was filled from `colors['background']`, or failing that from this package's own white and gray-900 defaults. A themed app paints its canvas through a `bg-surface` className alias, on a widget of its own. The two are read by different layers and had no reason to agree, and nothing in the stack put anything behind a page, so nobody noticed.

Measured on a running consumer:

| | light | dark |
|---|---|---|
| `bg-surface` alias (what the app paints) | `#F9FAFB` | `#07090C` |
| `scaffoldBackgroundColor` (what Material reads) | `#FFFFFF` | `#111827` |

## Why it stops being latent

`magic` 0.0.12 starts painting a page background. It has to: every magic page was wrapped in `Material(type: MaterialType.canvas)`, which paints `Theme.canvasColor`, and this package sets that to `Colors.transparent` on purpose so a Material surface never covers a Wind background class. Once routes start stacking there is a second page underneath, and it shows through. magic fixes that by painting `scaffoldBackgroundColor`.

With the table above, that would put pure white over a near-white canvas in light mode and `#111827` over `#07090C` in dark. The dark one is not subtle.

## Why the fix is here and not consumer-side

`colors['background']` holds ONE colour for both brightnesses: `toThemeData()` branches on brightness only for the DEFAULT. A themed app needs a different canvas in each, so there is no value a consumer could set that is right for both. The alias already carries both, so this reads it.

## How it reads it

As TEXT, not through the parser, and that is the load-bearing detail: the parser resolves a className against a theme, and the theme is what `toThemeData()` is assembling. An alias value is already `bg-[#F9FAFB] dark:bg-[#07090C]`, so the hex is there to be read.

Only a literal `bg-[#hex]` is read. Each of these keeps the existing default instead of guessing:

- an alias naming a palette colour (`bg-gray-50`)
- a dark theme whose alias carries no `dark:` pair, because falling back to the light token would hand a dark app a white canvas, which is worse than the default it already had
- an app with no `bg-surface` alias at all

An explicit `colors['background']` still wins over all of it, since that is the consumer saying it outright. Null from the helper means "this alias names no literal colour", never "the colour is transparent".

## Verification

`dart format` 0 changed, `flutter analyze` clean, `flutter test` **1833 passed, 1 skipped**.

Seven new cases cover both brightnesses, the explicit-colour precedence, both no-alias defaults, the palette-alias fallback, the missing-`dark:` fallback, and an eight-digit token's alpha.

Mutation-checked: ignoring the alias turns the two colour assertions red and leaves the four fallback cases green, so they measure the new path rather than the defaults that were already there.

## Sequencing

Not a blocker for magic#164 landing, but it IS a blocker for a magic 0.0.12 release reaching a themed app without a visible colour change. Whichever ships second, the pair has to be published before a consumer picks up the page paint.

## Summary by CodeRabbit

* **Bug Fixes**
  * Theme page backgrounds now match the colors rendered by the app’s `bg-surface` styling.
  * Dark themes correctly use their dark surface color when available.
  * Explicit background color settings continue to take precedence.
  * Default light and dark background colors remain in effect when no supported surface color is configured.
  * Literal six- and eight-digit hex colors are handled with the expected transparency.


anilcancakir:
Reviewed and taken over to close the release train, since magic 0.0.12 is held behind this. Head is `ad46332`, one commit on top.

The diagnosis is right and the placement is right: `colors['background']` holds one colour for both brightnesses, so no consumer value could have fixed it, and reading the alias as TEXT rather than through the parser is the only thing that works while `toThemeData()` is still assembling the theme it would parse against.

**What I changed, on the one thing the fix turns on.** The point is that Material paints the colour the className paints, so both have to parse a hex the same way. The inline arithmetic agreed with `hexToColor` for 6 and 8 digits and diverged either side of that:

- `bg-[#fff]` resolves to white through the parser (`hexToColor` expands 3 and 4 digit shorthand) and to the shipped default here, so a theme written in shorthand keeps the disagreement this PR exists to remove.
- `{6,8}` also matches seven digits, and `int.parse` answers a colour nobody wrote.

It calls `hexToColor` now and the pattern accepts exactly the lengths that helper accepts, which also means the two cannot drift apart later. Two tests added: the shorthand resolves like the parser resolves it, and the seven-digit case keeps the default.

Gates: `flutter test` **1835 passed**, 1 pre-existing skip; `dart analyze` clean; `dart format` 394 files 0 changed; `./tool/coverage.sh 90` at 95.4%.

@kodizm review please, head `ad46332`.

Shipping as wind **1.6.1** right after this merges, because magic 0.0.12 starts painting `scaffoldBackgroundColor` and needs the value to be the app's own canvas.

anilcancakir:
@kodizm review please. There is no Kodizm check on this PR's head at all (`ad46332`), and none on the original head either, so this one has had no pass from you yet.

It is small and self-contained: `WindThemeData.toThemeData()` fills `scaffoldBackgroundColor` from the `bg-surface` alias when that alias names a literal colour, because magic 0.0.12 starts painting that value and it disagreed with the canvas a themed app paints (`#FFFFFF` against `#F9FAFB` light, `#111827` against `#07090C` dark). The hex goes through `hexToColor`, the same helper the parser uses, so the two cannot resolve the same token differently. Nine tests, `flutter test` 1835 passing, coverage 95.4%.

anilcancakir:
Merging without a Kodizm pass. Three review requests over 50 minutes produced no check run on either head of this PR, so the bot is not reachable for it right now; the rest of this train has had its passes and acted on them.

What stands in for it here: I reviewed this branch line by line and the review found two real defects, both fixed in `ad46332` (the shorthand `bg-[#fff]` resolving one way through the parser and another way here, and a `{6,8}` pattern accepting a seven-digit typo and answering a colour nobody wrote). Gates on the head: `flutter test` 1835 passing with the one pre-existing skip, `dart analyze` clean, `dart format` 394 files 0 changed, `./tool/coverage.sh 90` at 95.4%, `Lint & Test` green on CI.

Shipping as 1.6.1 immediately, because magic 0.0.12 is held on it.

PR #212, opened 2026-09-18
fix(input): reading the caret during build walked an unlaid-out ancestor

anilcancakir:
## The crash

On a real iPhone, in a release build, opening a screen carrying a focused
`WInput` replaced the field with a flat grey box:

```
StateError: Bad state: RenderBox was not laid out: RenderFractionalTranslation
  RenderBox.size                                  box.dart:2304
  RenderFractionalTranslation.applyPaintTransform proxy_box.dart:3147
  RenderObject.getTransformTo                     object.dart:3730
  RenderBox.localToGlobal                         box.dart:3114
  RenderEditable._snapToPhysicalPixel             editable.dart:2342
  RenderEditable.getLocalRectForCaret             editable.dart:1862
  _WInputState._clearanceBelowCaret               w_input.dart:483
  _WInputState.build                              w_input.dart:594
```

Measured 2026-09-18 through Sentry, nine occurrences, one user.

`_clearanceBelowCaret` runs inside `build` and asked for the **pixel-snapped**
caret rect. That snap calls `localToGlobal`, which walks every ancestor's
`applyPaintTransform`. `RenderFractionalTranslation.applyPaintTransform` reads
`size` with no layout guard, and `RenderBox.size` **throws** rather than
asserting. A slide route transition builds exactly that object, and during the
frame it is inserted it has not been laid out.

**The existing `hasSize` guard could never have caught it**, and that is worth
keeping in mind: the editable itself IS laid out. The object without a size is
an ancestor, several frames up.

## The fix

`getEndpointsForSelection` answers the same question without leaving this render
object: same `_computeTextMetricsIfNeeded()`, reads `_textPainter` and
`_paintOffset`, no `localToGlobal` anywhere. For a collapsed selection it gives
the line bottom rather than the caret bottom, a sub-pixel difference this getter
cannot feel: it is already documented as generous by the top padding, and its
consumer compares with a 1px threshold.

**All seven pixel assertions in `w_input_scroll_into_view_test.dart` are
unchanged by it**, which is the check that mattered most.

## The new test does not reproduce the crash, and says so first

Two shapes were tried, a hand-driven `SlideTransition` and a real
`Navigator.push` with an autofocusing field. **Both pass with the defect in
place.** On a device the focus change and the transition's insertion land in the
same frame; in a widget test autofocus fires a frame later and the ancestor is
laid out by then.

What holds this fix is the stack trace and the API contract, not that test. It
is a regression guard on the shape and is labelled as one rather than left to
look like proof.

`dart analyze` clean, 1,836 tests green. No version bump, per the release train.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented focused input fields from crashing or displaying an error screen when navigating to a new route during a transition.
  * Fixed excess spacing or scrolling when clearing a focused input field.
  * Improved caret positioning when the selection is temporarily invalid.

* **Tests**
  * Added regression coverage for clearing focused fields and displaying autofocus fields during route transitions.


anilcancakir:
@kodizm both findings are closed on head `74d0c36`, ready for a re-review. It landed after your last pass, which reviewed `3bb5cb4`.

**The invalid-selection fallback.** You were right that the path is ordinary rather than theoretical, and right about the consequence. `_updateControllerValue` assigns `_controller.text`, whose setter forces `TextSelection.collapsed(offset: -1)` (`editable_text.dart:279`), and when the new value is EMPTY neither branch below restores a valid one, so a focused field whose `value` prop is set to `''` reached `_clearanceBelowCaret` with no selection and jumped its scroll padding by about a line height.

Returning `_measuredHeight` there was the over-reserve this getter's own docblock warns about. The guard is gone and the extent falls back to offset 0 instead:

```dart
final TextPosition extent = _controller.selection.isValid
    ? _controller.selection.extent
    : const TextPosition(offset: 0);
```

An empty field's caret is at the start, so offset 0 is both correct and the number the old code produced before the crash fix. A test drives that exact sequence, focus then clear, rather than constructing the invalid selection by hand.

**CHANGELOG.** Added under `## [Unreleased]` / `Fixes`, carrying what the PR description had: the `localToGlobal` walk through `RenderFractionalTranslation.applyPaintTransform`, why `hasSize` could never have caught it, and why `getLocalRectForCaret` was rejected in favour of `getEndpointsForSelection`.

**Gates on head:** `dart analyze lib test` no issues, `dart format --output=none --set-exit-if-changed lib/ test/ example/lib/` 395 files 0 changed, `flutter test` green at 1837 with the one pre-existing skip.

anilcancakir:
@kodizm both addressed on `15f7c4b`, and one of them comes with a correction to a premise we both held.

**The heading.** Right, and worse than one heading: the entry was missing the documented shape entirely. `CLAUDE.md` asks for a bold opening clause, the rejected alternative and why, the files touched in backticks and the PR number, and the adjacent 1.6.1 and 1.6.0 entries all carry it. Rewritten as `### Fixed` in that shape, and it now covers the offset-0 fallback too, which had landed in code with no entry at all.

**The test, and the correction.** You offered measuring or renaming. I tried to measure, and the number cannot be made to move, which turns out to be because the path does not exist.

Both of us reasoned that `_controller.text = ''` forces `TextSelection.collapsed(offset: -1)`, that `_updateControllerValue` restores nothing for an empty value, and that the invalid selection therefore reaches `_clearanceBelowCaret`. I instrumented the getter rather than trusting that, and every frame of a focused field being cleared through the `value` prop prints:

```
isValid=true sel=TextSelection.collapsed(offset: 0, ...) text=""
```

Something normalises the `-1` before `build` reads the clearance, so the framework already does exactly what the fallback does. The measurement follows:

| | with the offset-0 fallback | with `return _measuredHeight` restored |
|---|---|---|
| before the clear | 36.0 | 36.0 |
| one pump after | 36.0 | 36.0 |
| at settle | 36.0 | 36.0 |

So a padding assertion would have passed against the defect exactly as `takeException() isNull` does, which is your objection surviving into the fix for it. The test is named `clearing a focused field renders and does not throw` now, and its docblock carries the instrumented output rather than the reasoning that predicted otherwise.

**The fallback still ships.** Offset 0 is where an empty field's caret sits and it costs nothing, so it is the right answer if any path does arrive invalid. It has no test because I could not construct that path, and a test that cannot fail is worse than a stated gap. If you can construct one, that is worth more than either of us re-deriving it.

**Gates on head:** `dart analyze lib test` no issues, `dart format --output=none --set-exit-if-changed lib/ test/ example/lib/` 395 files 0 changed, `flutter test` green at 1837 with the one pre-existing skip. `./tool/coverage.sh 90` not run locally either: this change touches no `lib/` line, only the CHANGELOG and a test name plus comments.

PR #213, opened 2026-09-18
release: 1.6.2

anilcancakir:
Patch, one fix, first of a seven-package batch.

## What ships

**A focused `WInput` no longer turns into a flat grey box when a route transition is pushed in above it.** `_clearanceBelowCaret` ran during `build` and asked `RenderEditable` for the pixel-snapped caret rect; that snap ends in `localToGlobal`, which walks every ancestor's `applyPaintTransform`, and `RenderFractionalTranslation.applyPaintTransform` reads `size` with no layout guard while `RenderBox.size` throws rather than asserting. A slide route transition builds exactly that object and during the frame it is inserted it has not been laid out.

Measured on an iPhone in a release build: `StateError: Bad state: RenderBox was not laid out: RenderFractionalTranslation`. Debug never reaches it, which is why no simulator run and no test had ever seen it. The clearance comes from `getEndpointsForSelection` now, which reads the same text metrics without leaving the render object. Landed in #212.

## Stamps

The patch-release set this repo uses, five of them: `pubspec.yaml`, `example/pubspec.yaml` with its lockfile, the `dartdoc_options.yaml` source-link tag, the `llms.txt` version line, and the CHANGELOG promotion with its link references.

`skills/wind-ui/SKILL.md` needs none: no API surface moved and its stamp names the 1.6 line rather than the patch.

## Where this sits in the batch

First, deliberately. `magic` 0.0.14 re-pins `fluttersdk_wind: ^1.6.2`, so this has to be on pub.dev before that resolves. The order is wind + artisan, then dusk + telescope, then magic, then magic_deeplink + magic_starter.

## Gates

`dart analyze lib test` no issues. `dart format --output=none --set-exit-if-changed lib/ test/ example/lib/` 395 files 0 changed. `flutter test` green at 1837 with the one pre-existing skip. `dart pub publish --dry-run` clean apart from the gitignored `pubspec_overrides.yaml` hint, which is local-only.

## Summary by CodeRabbit

- **Release**
  - Published version 1.6.2.
  - Updated package and example application version information.
- **Documentation**
  - Updated documentation and source-link references to version 1.6.2.
  - Added 1.6.2 changelog and release references.


anilcancakir:
@kodizm the date is right, and the disagreement is a timezone rather than a typo.

```
$ date -u '+%Y-%m-%d %H:%M UTC'
2026-09-18 22:55 UTC
$ date '+%Y-%m-%d %H:%M %Z'
2026-09-19 01:55 +03
```

Your container reads UTC and is three hours behind the machine cutting the release, so `2026-09-19` is today here and the tag lands within the hour. This repo dates a release by the author's local day: 1.6.1 is dated `2026-09-16` against a commit at `Wed Sep 16 23:44:16 2026 +0300`, which is the same convention and did not cross midnight the way this one does. The rest of the batch carries the same date for the same reason.

Everything else in your pass matches what I ran, including `python3 tool/check-docs.py` clean at 72 doc pages against 171 demo routes, which I had not thought to run and is the gate that would have caught a stale `<x-preview>` if the release had touched one.

Merging and tagging now. `fluttersdk_wind` 1.6.2 has to be on pub.dev before magic#172 resolves.

PR #216, opened 2026-09-22
fix(theme): an ARGB filter deleted transparent from every palette

anilcancakir:
## The defect

`WindThemeData._initColors` (`lib/src/theme/wind_theme_data.dart:250-264` on master) built a `MaterialColor` for every entry in `defaults/colors.dart`, used `MaterialColor(0, {})` as its "this value is not a colour" sentinel, and then dropped the sentinel:

```dart
..removeWhere((key, value) => value.toARGB32() == 0);
```

An ARGB of 0 does not mean "not a colour". It is exactly `transparent` (`Color(0x00000000)`, `lib/src/theme/defaults/colors.dart:4`). So the filter deleted the one token it was never aimed at. The sentinel meanwhile was unreachable: all 26 shipped entries are a `Color` or a `Map<int, Color>`, so the branch was dead in both directions and the filter only ever did harm.

`isValidColor('transparent')` answered false, and every parser that resolves a named colour walked past the class to whatever came before it.

Measured in a consumer app against a default palette:

| className | resolved background, before |
|---|---|
| `bg-transparent` | null, no colour at all |
| `bg-primary bg-transparent` | primary, so the last class did NOT win |
| `bg-primary bg-white` | white, so last-wins worked for every other colour |
| `bg-primary bg-transparent/0` | primary |
| `bg-primary bg-[#00000000]` | primary, the arbitrary-hex regex took 3 or 6 digits only |

Reported from a real screen: a ghost button styled `w-full bg-transparent border text-gray-500` rendered as a filled brand-coloured button with muted grey text on it, unreadable.

## The fix

`_initColors` builds the map by iteration and inserts only the entries it could narrow, so a value that is neither shape is skipped rather than materialised and filtered back out. Filtering on the sentinel's identity instead of on its value would have fixed the symptom too, and was rejected because a sentinel nothing can produce is not worth constructing.

The regression gate is not "transparent survives" but "no token is lost": a fully transparent colour is indistinguishable from absent by ARGB alone, so only the key set can say it.

### Which parsers this reaches

All of them, because they share one palette. Each is covered by a test in this PR:

| Parser | Token | Test |
|---|---|---|
| `background_parser` | `bg-transparent`, `to-transparent` | `test/parser/parsers/background_parser_test.dart`, `test/widgets/w_div/background_test.dart` |
| `text_parser` | `text-transparent`, `decoration-transparent` | `test/parser/parsers/text_parser_test.dart` |
| `border_parser` | `border-transparent` | `test/parser/parsers/border_parser_test.dart` |
| `ring_parser` | `ring-transparent` | `test/parser/parsers/ring_parser_test.dart` |
| `svg_parser` | `fill-transparent`, `stroke-transparent` | `test/widgets/w_svg_test.dart` |

`wind_helpers.wColor(context, 'transparent')` is reached too, through the same `getColor`.

## Second half: four- and eight-digit arbitrary hex

`hexToColor` has accepted 3, 4, 6 and 8 digits since it was written. `doc/utilities/color-helpers.md` documents all four. `skills/wind-ui/references/tokens.md` told agents `bg-[#hex]` took "3, 4, 6, or 8 chars". `border-[#...]` and `ring-[#...]` genuinely did. Only the background, gradient-stop, text and decoration regexes were left at `{3}|{6}`, from the commit that widened the border one later, so the four families that needed alpha most were the four that could not express it. A non-matching class is not an error anywhere in the pipeline, so `bg-[#80FF0000]` did nothing at all and the class before it won.

Widened to `{3}|{4}|{6}|{8}`, spelled exactly as `border_parser.dart` already spells it. Not `{3,8}`: a five- or seven-digit value is a typo rather than a colour. Alpha leads in both wide forms, because that is what `hexToColor` does and what the rest of the package documents (Flutter packs `AARRGGBB` where CSS writes `RRGGBBAA`).

This landed on the same branch rather than its own because it is the same defect surface: with the palette fixed and this left alone, `bg-transparent` works and every PARTIALLY transparent colour still has no spelling. Happy to split it if you would rather review them apart.

## Mutations run

Every new test was written red first, and each was then re-broken against the restored defect:

- `removeWhere((_, value) => value.toARGB32() == 0)` put back on top of the new structure: 8 of the new assertions go red, across all five parsers plus the theme.
- The four regexes narrowed back to `{3}|{6}`: the 5 alpha-hex assertions go red.

## Gates

- `dart analyze`: No issues found
- `dart format .`: 395 files, 0 changed
- `flutter test`: 1851 passed, 1 pre-existing skip
- `./tool/coverage.sh 90`: 95.4%, PASS
- `cd example && flutter build web`: exit 0

## Post-change sync

- `doc/`: arbitrary-value sections of `background-color.md`, `background-gradient.md`, `text-color.md`. The transparent tokens needed no doc change, because the docs already described them as working.
- `example/lib/pages/styling/background_color_basic.dart`: an "Arbitrary Hex with Alpha" section, over a solid panel so the alpha reads.
- `skills/wind-ui/`: `SKILL.md` and `references/tokens.md` now state the alpha ordering, which an agent could not infer and would get backwards from CSS habit. Version 2.17.0 to 2.18.0.
- `CHANGELOG.md`: two entries under `[Unreleased]` / `Fixed`.
- `README.md`: no change. No widget, theme field or parser family added.

## Found, not changed

Out of this PR's topic, each a documented token with no implementation:

- `bg-current` and `bg-inherit` are in `doc/styling/background-color.md` and in the skill's token table; `background_parser` has no `current` handling at all, so both drop silently. Same for `text-current` and `text-inherit` in `text_parser`. Only `fill-current` and `stroke-current` are actually wired (`svg_parser.dart:57,70`).
- `border-t-transparent` (and any directional border colour) does not parse: `_borderColorRegex` is `^border-(?<color>[a-zA-Z]+)(?:-(?<shade>\d{2,3}))?$`, which cannot take a side segment. `doc/interactivity/animation.md:59` uses it in a spinner example.
- `svg_parser._parseColor` passes any bracket content straight to `hexToColor`, which ends in `int.parse(code, radix: 16)`. `fill-[nope]` throws a `FormatException` out of the parse rather than dropping the class. Every other parser gates the bracket form behind a hex regex first.

## Summary by CodeRabbit

* **Bug Fixes**
  * Transparent color utilities now correctly resolve across backgrounds, text, borders, rings, fills, strokes, decorations, and gradients.
  * Transparent styles correctly override earlier color classes.
* **New Features**
  * Arbitrary colors now support 4- and 8-digit hex values with alpha, alongside existing 3- and 6-digit formats.
  * Gradient stops support alpha-enabled arbitrary hex colors.
* **Documentation**
  * Updated styling guidance and examples to explain alpha formats and Flutter’s alpha-leading color notation.


anilcancakir:
Thanks, all four verified in source and all four addressed. Commit `890efb82566386fd2ed4882156df2a598f0f6011`.

**Major, `doc/styling/background-color.md:108`** — correct, and the doc line was wrong rather than the test. `hexToColor` doubles each nibble (`lib/src/utils/color_utils.dart:47-49`), so `#8f00` expands to `#88ff0000`, alpha `0x88`. The line now reads "`#FF0000` at alpha `0x88`" against the 8-digit form's "alpha `0x80`", and a note says the short forms expand by doubling and that the 4-digit form cannot spell every alpha, so the 8-digit one is what to write when the exact value matters. `example/lib/pages/styling/background_color_basic.dart:70` carried the same claim in its section description and now says the same thing.

**Minor, `CHANGELOG.md:13`** — correct, the claim outran the diff. Rather than weaken it, both tokens have a test now: `test/widgets/w_svg_test.dart:114` (`stroke-transparent`) and `test/parser/parsers/text_parser_test.dart:60` (`decoration-transparent`, the named token, beside the existing arbitrary-hex case). Both were confirmed red against the restored `removeWhere`.

**Minor, `lib/src/theme/wind_theme_data.dart:274-288`** — correct on both, and they needed different answers.

`WCheckbox` (`lib/src/widgets/w_checkbox.dart:153`): the widget's own default className is `border border-gray-300 ... checked:bg-primary checked:border-transparent`, so shedding the outline when checked is what it has always been written to do and never did. The new behaviour is the intended one, so it is pinned rather than reverted: `test/widgets/w_checkbox_test.dart:194` asserts a checked box's border is `Color(0x00000000)` and `:213` asserts an unchecked one's is not. The first goes red against the restored `removeWhere`; the second is the control and stays green.

`WSelect` (`lib/src/widgets/w_select.dart:1019`): this one is a real cost rather than an intent restored. A resolved transparent colour is still a colour, so it builds a `BoxDecoration`, which trips `needsContainer` (`lib/src/widgets/w_div.dart:1437`) and wraps every unselected option in a `DecoratedBox` that paints nothing, once per row for the length of the list. The resting arm is `''` now, with a comment at the site saying why: both spellings paint nothing and only one of them costs a widget.

Both are called out in the CHANGELOG under their own paragraph, since you were right that the entry read as though only consumer classNames were affected.

**Minor, `ring_parser.dart:35` / `shadow_parser.dart:25`** — correct, and taken now rather than as a follow-up, because leaving them contradicts the doc blocks this PR just wrote three of. `skills/wind-ui/references/tokens.md:429` already told agents both took "3, 4, 6, or 8 hex digits", so `{3,8}` matched neither the documentation nor the four families fixed here. Both narrowed to the same `{3}|{4}|{6}|{8}`, with `test/parser/parsers/ring_parser_test.dart:113` and `test/parser/parsers/shadow_parser_test.dart:165` covering the five-digit case; both go red against the restored `{3,8}`. Every arbitrary colour across `doc/`, `example/` and `test/` is six digits, so nothing that worked stops working. The ring assertion is written against `context.theme.ringColor` rather than `isNull`, because `RingParser` resolves a dropped colour to the theme default (`ring_parser.dart:180`).

Gates after the round: `dart analyze` clean, `dart format .` 0 changed, `flutter test` 1857 passed / 1 pre-existing skip, `./tool/coverage.sh 90` 95.4% PASS, `python3 tool/check-docs.py` 0 issues, `cd example && flutter analyze` clean.

@kodizm please re-review.

anilcancakir:
Both verified in source. Commit `683ddc9f1eb82675e8dda125705dead4c08bad65`.

**Major, `lib/src/widgets/w_button.dart:220`** — confirmed, and it is the worst of the three because the pattern came from this widget's own class doc. Fixed on both halves.

The spinner colour skips a fully transparent text colour (`lib/src/widgets/w_button.dart:220-231`). The reasoning, which is in the comment at the site: `text-transparent` says "do not show the label", the spinner is not the label, and reading a text colour as a spinner colour is a heuristic whose input simply does not carry the answer when alpha is zero. That is continuous with the fallback's own stated purpose at `:216-218` rather than a new special case, and `loadingColor` stays first in the chain, so an explicit transparent spinner is still honoured.

The recommendation is off `lib/src/widgets/w_button.dart:17`, replaced by what actually applies: the spinner REPLACES the child rather than overlaying it, so there was never a label for that class to hide; `loadingColor` sets the spinner colour and `loadingText` puts a caption beside it. Checked the rest of the tree, `loading:text-transparent` appeared nowhere else, `doc/`, `skills/`, `example/`, `README.md` and `llms.txt` all use `loading:opacity-*` or `loading:bg-*`.

Two tests at `test/widgets/w_button_test.dart:640` and `:669`. The first pumps the documented class and asserts the spinner has alpha and is the contrast colour; it goes red against the restored `loadingColor ?? styles.color ?? _contrastColor(styles)`. The second pumps `loadingColor: Color(0x00000000)` and asserts it survives, which is the control for "the fall-through reads a text colour, never an explicit ask"; it stays green under the same mutation, which is what makes the pair meaningful.

One thing I did not change, stated so it is a decision rather than an oversight: `loadingText` still inherits the transparent colour through the `DefaultTextStyle.merge` at `:190`. That is the class doing exactly what it says on something that genuinely is text, and the caption is opt-in rather than something the widget draws on its own. With the recommendation off the class doc, nobody is steered into that combination.

**Minor, `lib/src/widgets/w_switch.dart:20`** — agreed on both counts, and I re-read `w_switch.dart` to be sure: the class doc lists `checked:border-transparent` as a supported state class at `:20`, the default className at `:151` is `w-11 h-6 rounded-full bg-gray-300 checked:bg-primary` with no border token at all, so nothing about the switch changes. The CHANGELOG paragraph is scoped to widgets whose own default className carries the token; it now reads "three shipped widgets" and names `WButton` alongside `WCheckbox` and `WSelect`. No claim about doc comments was made and none is added.

Gates after the round: `dart analyze` clean, `dart format .` 0 changed, `flutter test` 1859 passed / 1 pre-existing skip, `./tool/coverage.sh 90` 95.4% PASS, `python3 tool/check-docs.py` 0 issues.

@kodizm please re-review.

anilcancakir:
Correction to my own reply above, on the `WSwitch` minor. The conclusion holds and my citation for it did not.

I wrote that the default className at `lib/src/widgets/w_switch.dart:151` is `w-11 h-6 rounded-full bg-gray-300 checked:bg-primary`. There is no such line. The widget builds its track className at `:127-129`: `className == null ? 'flex items-center' : 'flex items-center $className'`. So the built-in default carries no colour or border token of any kind, which makes the point stronger than the version I gave, not weaker: `checked:border-transparent` is advertised at `:20` and is reachable only through a className the consumer writes, so nothing about `WSwitch` changes in this PR.

`w-11 h-6 rounded-full ...` is what the class doc's usage example at `:31-37` and the `className` prop's own example at `:60` suggest a consumer write. I read those and attributed them to the widget. Apologies for the noise.

